PR #3, opened 2018-08-30
WIP Kotlin tests MySQL

dougEfresh:


db-async-common/src/main/java/com/github/jasync/sql/db/util/ExecutorServiceUtils.kt
```diff
@@ -10,7 +10,7 @@ object ExecutorServiceUtils {
   //creating a lot of threads from that pool means we are not really reactive
   val CachedThreadPool: ExecutorService = Executors.newCachedThreadPool(DaemonThreadsFactory("db-sql-default"))
   val CommonPool: ExecutorService = ForkJoinPool.commonPool()
-  val CommonPoolNetty: ExecutorService = io.netty.util.internal.chmv8.ForkJoinPool.commonPool()
```
dougEfresh: package was removed in latest version of netty

oshai: We should probably upgrade netty to latest

mysql-async/src/test/java/com/github/jasync/sql/db/mysql/BinaryColumnsSpec.kt
```diff
@@ -0,0 +1,149 @@
...
+        withConnection {
+            connection ->
+            executeQuery(connection, create)
+            executePreparedStatement(connection, insert, listOf(1, bytes))
+            executePreparedStatement(connection, insert, listOf(2, ByteBuffer.wrap(bytes.toByteArray())))
+            executePreparedStatement(connection, insert, listOf(3, Unpooled.wrappedBuffer(bytes.toByteArray())))
+
+            val rows = executeQuery(connection, select).rows
+            assertEquals(3, rows?.size)
+            assertEquals(rows?.get(0)?.get("id"), 1)
```
oshai: the expected should be first - will give better errors in case of failure

mysql-async/src/test/java/com/github/jasync/sql/db/mysql/BinaryColumnsSpec.kt
```diff
@@ -0,0 +1,149 @@
...
+            assertEquals(rows?.get(1)?.get("id"), 2)
+            assertEquals(rows?.get(1)?.get("blob_column"), bytes)
+            assertEquals(rows?.get(2)?.get("id"), 3)
+            assertEquals(rows?.get(2)?.get("blob_column"), bytes)
+        }
+
+    }
+
+    fun compareBytes(row: RowData, column: String, expected: String) {
+        assertEquals(row[column], expected.toByteArray(CharsetUtil.UTF_8))
```
oshai: expected should be first - see comment above

mysql-async/src/test/java/com/github/jasync/sql/db/mysql/BitSpec.kt
```diff
@@ -0,0 +1,68 @@
...
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+
+class BitSpec : ConnectionHelper() {
+
+
+    @Test
+    fun `result in binary data` () {
+
+      withConnection {
```
oshai: In tests that use a connection I suggest to do something like:
` fun `result in binary data` () = withConnection { connection -> ...`
It will reduce nesting level and keep the boilerplate out

db-async-common/src/main/java/com/github/jasync/sql/db/util/AbstractURIParser.kt
```diff
@@ -107,8 +107,8 @@ abstract class AbstractURIParser {
         val userInfo = parseUserInfo(uri.userInfo)
 
         val port = uri.port.nullableFilter { it > 0 }
-        val db = uri.path.nullableMap { it.removeSuffix("/") }.nullableFilter { !it.isEmpty() }
-        val host = uri.host
+        val db = uri.path.nullableMap { it.removePrefix("/") }.nullableFilter { !it.isEmpty() }
```
oshai: good catch!

db-async-common/src/main/java/com/github/jasync/sql/db/util/AbstractURIParser.kt
```diff
@@ -107,8 +107,8 @@ abstract class AbstractURIParser {
...
+        val host = uri.host.nullableMap { it.removePrefix("[").removeSuffix("]") }
```
oshai: was that in the original code?

db-async-common/src/main/java/com/github/jasync/sql/db/util/ExecutorServiceUtils.kt
```diff
@@ -10,7 +10,7 @@ object ExecutorServiceUtils {
...
+  val CommonPoolNetty: ExecutorService = ForkJoinPool.commonPool()
```
oshai: we can mark it as deprecated/ remove it.

mysql-async/src/test/java/com/github/jasync/sql/db/mysql/ContainerHelper.java
```diff
@@ -0,0 +1,44 @@
...
+        "localhost",
+        33306,
+        "root",
+        "mysql_async_tests");
+
+  static {
+    try {
+      new MySQLConnection(defaultConfiguration).connect().get(1, TimeUnit.SECONDS);
+    } catch (Exception e) {
+      // If local instance isn't running, start a docker mysql on random port
```
oshai: maybe print it to console?

mysql-async/src/test/java/com/github/jasync/sql/db/mysql/codec/MySQLFrameDecoderSpec.kt
```diff
@@ -1,59 +1,33 @@
...
+  fun `decode an OK message correctly` () {
       val buffer = createOkPacket()
-
       val decoder = this.createPipeline()
-
       decoder.writeInbound(buffer)
 
       val ok = decoder.readInbound().asInstanceOf[OkMessage]
-      ok.affectedRows === 10
+      traok.affectedRows === 10
```
oshai: typo?

mysql-async/src/test/java/com/github/jasync/sql/db/mysql/BinaryColumnsSpec.kt
```diff
@@ -8,20 +8,21 @@ import java.util.*
...
 
         val create = """CREATE TEMPORARY TABLE t (
-                     |  id BIGINT NOT NULL AUTO_INCREMENT,
-                     |  uuid BINARY(36) NOT NULL,
-                     |  address VARBINARY(16) NOT NULL,
-                     |  PRIMARY KEY (id),
-                     |  INDEX idx_t_uuid (uuid),
-                     |  INDEX idx_t_address (address)
-                     |);"""
+                       id BIGINT NOT NULL AUTO_INCREMENT,
```
oshai: I think you should leave the pipe here and use trim indent: 
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/trim-margin.html

oshai:
It seems like a lot of work, Is there something I can help with?

dougEfresh:
@oshai Yeah, I need help with some test failures.
It seems we are using java.time.Duration in the Decoder/Encoder (see TimeDecoder )
Do we want to use java's time lib or Joda's (I prefer java.time).

oshai:
From the code, I think the original lib used joda so it might be less risky to move. On the other hand, it is better to have fewer dependencies 
So let's go with `java.time` 

You can merge even if some tests are failing and then it will be easier for me to help, just let me know what to take over.

dougEfresh:
@oshai 
The duration tests are failing, lets fix those 1st. I'm not sure why they fail. 
FYI, I am using https://github.com/testcontainers/testcontainers-java 
Do you don't need to run any special script for a local mysql instance, however, to speed up the tests, you can run the run-docker-mysql.sh script.

PR #8, opened 2018-09-05
Adding transaction test

dougEfresh:
Look at the test I skip...
adding 
``` 
 val r = awaitFuture(future)
            assertEquals(1, r.rowsAffected)
```
passes the test...which doesn't make sense to me.


oshai:
Why not? you have 2 insert queries, each one affect one row.

mysql-async/src/test/java/com/github/jasync/sql/db/mysql/TransactionSpec.kt
```diff
@@ -1,25 +1,49 @@
...
-//import kotlin.test.assertEquals
-//import kotlin.test.assertNotNull
+import com.github.jasync.sql.db.util.ExecutorServiceUtils
+import com.github.jasync.sql.db.util.flatMap
+import org.junit.Ignore
+import org.junit.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
 
 class TransactionTest : ConnectionHelper() {
```
oshai: The name of the class does not match the name of the file, which is confusing.

PR #24, opened 2018-10-10
add typed access to RowData

oshai:


oshai:
see wiki: https://github.com/jasync-sql/jasync-sql/wiki/Type-Safety

PR #26, opened 2018-10-16
Actor based object pool

oshai:
This pull request introduces a new type of object pool which uses coroutines Actor.
Notable API changes:
A couple of configurations were moved from `Configuration` to `PoolConfiguration`:
* createTimeout
* testTimeout

This was changed because in the new implementation pool is responsible for those actions.
The old configuration is now deprecated, but the usage of the new/old configuration is based on the used pool.

Eventually, I plan to make that ActorBased pool the default and only implementation.

I will merge it after Kotlin 1.3 is out with coroutines not in experimental mode, in order to prevent compatibility issues.


oshai:
The branch was merged directly to master.

PR #27, opened 2018-10-17
downgrade netty to 4.0.56

oshai:


oshai:
created version `0.8.37-netty4.0` with it

dougEfresh:
@oshai Look at my latest commit to your branch. I removed netty-all as it contains a lot of crap we don't need. Also, I added a travis test for netty-4.1.30.Final....so we are testing both netty versions.

PR #34, opened 2018-11-12
add a scope for the object pool actor

oshai:
so we can cancel it later when  pool is closed


oshai:
created a new PR for this

PR #39, opened 2018-11-15
add fix for inactive channel bug

oshai:
Found that in mysql proxy sometimes immediately after query is sent
there is a channel inactive message and no more messages
Even timeout on connection is not working because of a check if connection is active in disconenct method

Add 2 fixes:
- In case of timeout, fail the query future even if not connected
It happens only once anyway
- Add a query timeout parameter to the pool, which is observing all items from outside anyway


dougEfresh:
This seems like a bug with MySQL proxy as well

oshai:
Might be, but it happens only once in couple of hours, so I'm not sure how severe it is without this issue.

PR #52, opened 2018-12-10
Add connection timeout to the channel to release the resource timely

bytewayio:
Add connection timeout to the source socket to release the socket resource as soon as the connection timeout

### DIFF
--- a/db-async-common/src/main/java/com/github/jasync/sql/db/Configuration.kt
+++ b/db-async-common/src/main/java/com/github/jasync/sql/db/Configuration.kt
@@ -38,5 +38,6 @@ data class Configuration @JvmOverloads constructor(
     val charset: Charset = CharsetUtil.UTF_8,
     val maximumMessageSize: Int = 16777216,
     val allocator: ByteBufAllocator = PooledByteBufAllocator.DEFAULT,
+    val connectionTimeout: Duration? = null,
     val queryTimeout: Duration? = null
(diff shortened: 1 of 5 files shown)


oshai:
@bytewayio - adding this looks like a good idea. a couple of comments:
* ConnectionPoolConfiguration already have a configuration for `connectionCreateTimeout` and I think that should be reused.
* We should probably add that to MySQL as well.
* I am not sure why you omitted the decoder.

bytewayio:
@oshai 
#1 in ConnectionPoolConfiguration, I copied the connection timeout setting to Configuration, I don't know what you meant by reuse the connectionCreateTimeout, but we should remove the connectionTimeout from pool

#2 MySQL is covered, see MySQLConnectionHandler

#3 because the decode is not used, no need to keep it

oshai:
@bytewayio re  comment 1: I don't think it is a good idea to remove it from pool configuration because it gives another safety on top of the connection configuration (helps to make sure pool is clean etc') so we can keep both. I suggest to make it non-nullable and of type Long (or Int) so it will be similar to the type in the pool.

The other points you're correct, I probably missed that.

bytewayio:
@oshai done with making the connectionTimeout as non-nullable and default to 5 seconds (some as the pool configuration)

oshai:
@bytewayio - thanks for the PR!

PR #54, opened 2018-12-10
Fix charset mapper

godpan:
CharsetMapper update see mysql doc:
[CharacterSet](https://dev.mysql.com/doc/internals/en/character-set.html#packet-Protocol::CharacterSet)


oshai:
I am not sure what is the problem that you fixed so if you can clear that up it will be great.

godpan:
When I analyze the mysql packet,the charset value is not same mysql doc make me confused，others will be think so. and in map have same key is not good idea.

oshai:
@godpan - I agree about the same key in the map, it was like that in the original lib, but I will take a look there. About the mapping itself, I am really not sure I fully understand the use case.

oshai:
@godpan - I did a partial change on #55 - let me know if that is enough.

godpan:
I am sorry, did not read the document carefully, 
they can all be expressed as utf-8,i will colse this pr.

PR #68, opened 2019-01-03
Epoll support

oshai:


postgresql-async/build.gradle
```diff
@@ -7,6 +7,7 @@ dependencies {
     compile "org.joda:joda-convert:$joda_convert_version"
     compile "io.netty:netty-transport:$netty_version"
     compile "io.netty:netty-handler:$netty_version"
+    compile "io.netty:netty-transport-native-epoll:$netty_version:linux-x86_64"
```
andy-yx-chen: Can you add kqueue support as well? for macOs/freeBSD

oshai: @andy-yx-chen I am not sure how usable that is (mac servers?). In addition, do you know if there is a way to check if it is supported (like Epoll.isAvailable())?

andy-yx-chen: we can cover freeBSD servers, it's KQueue.isAvailable()

oshai: @andy-yx-chen thanks, I added that as well

andy-yx-chen: @oshai , thanks, looks pretty good

PR #69, opened 2019-01-04
Infer SocketChannel type from EventLoopGroup type

andy-yx-chen:
This is to prevent Channel vs EventLoop mismatch, and allow customize EventLoopGroup instead of assuming EventLoopGroups will be always created from NettyUtils


oshai:
Thanks!

PR #89, opened 2019-01-31
checking log level trace on build

oshai:


oshai:
it removes false coverage on debug message and don't seem to affect build time.

PR #94, opened 2019-02-13
add thread pools to Configuration class instead of directly setting them

oshai:
add also validation on configuration


andy-yx-chen:
looks good, though a breaking change for vertx driver

oshai:
@andy-yx-chen I am sorry for the inconvenience. Did some breaking changes in 0.9 and missed that one. It is going to be released in 0.9.20

andy-yx-chen:
No worries, I will get rid of all compatibility issues

PR #96, opened 2019-02-21
Add maxTtl do pool objects

akleiman:
Hi there. We talked briefly yesterday (21/02) about some things I'd like to see in jasync. I had mentioned creating a PR to allow user-defined properties for postgres, and had mentioned the need to have a maxTtl for the pool. Shortly after we talked we hit a serious problem in production and maxTtl became much more urgent, so I prepared this PR. I'm willing to discuss and change any aspect of it, of course.

It was straightforward to implement and test; I chose to have two checks, one in `sendAvailableItemsToTest`, but also to have one in `validate` (so it gets triggered on `GiveBack` and `Take`).

The one thing I wasn't entirely sure about was default values; I chose to treat negative values of maxTtl as not having a TTL at all, as the previous behavior of the library, and chose to default to -1 if not specified, so as to not produce any surprising results on version upgrade.

### DIFF
--- a/db-async-common/src/main/java/com/github/jasync/sql/db/ConnectionPoolConfiguration.kt
+++ b/db-async-common/src/main/java/com/github/jasync/sql/db/ConnectionPoolConfiguration.kt
@@ -36,7 +36,7 @@ import java.util.function.Supplier
  * @param allocator the netty buffer allocator to be used
  *  * @param maxObjects how many objects this pool will hold
  * @param maxIdleTime number of milliseconds for which the objects are going to be kept as idle (not in use by clients of the pool)
- * @param maxTtl number of milliseconds an object in this pool should be kept alive, negative values mean no aging out
+ * @param maxConnectionTtl number of milliseconds an object in this pool should be kept alive, negative values mean no aging out
  * @param maxPendingQueries when there are no more objects, the pool can queue up requests to serve later then there
  *                     are objects available, this is the maximum number of enqueued requests
  * @param connectionValidationInterval pools will use this value as the timer period to validate idle objects.
@@ -57,7 +57,7 @@ data class ConnectionPoolConfiguration @JvmOverloads constructor(
     val username: String = "dbuser",
     val password: String? = null,
     val maxActiveConnections: Int = 1,
-    val maxTtl: Long = -1,
+    val maxConnectionTtl: Long? = null,
     val maxIdleTime: Long = TimeUnit.MINUTES.toMillis(1),
     val maxPendingQueries: Int = Int.MAX_VALUE,
     val connectionValidationInterval: Long = 5000,
@@ -85,6 +85,7 @@ data class ConnectionPoolConfiguration @JvmOverloads constructor(
         require(connectionCreateTimeout >= 0) { "connectionCreateTimeout should not be negative: $connectionCreateTimeout" }
         require(connectionTestTimeout >= 0) { "connectionTestTimeout should not be negative: $connectionTestTimeout" }
         queryTimeout?.let { require(it >= 0) { "queryTimeout should not be negative: $it" } }
+        maxConnectionTtl?.let { require(it >= 0) { "queryTimeout should not be negative: $it" } }
     }
 
     val connectionConfiguration = Configuration(
@@ -108,7 +109,7 @@ data class ConnectionPoolConfiguration @JvmOverloads constructor(
     val poolConfiguration = PoolConfiguration(
         maxObjects = maxActiveConnections,
         maxIdle = maxIdleTime,
-        maxTtl = maxTtl,
+        maxObjectTtl = maxConnectionTtl,
         maxQueueSize = maxPendingQueries,
         validationInterval = connectionValidationInterval,
         createTimeout = connectionCreateTimeout,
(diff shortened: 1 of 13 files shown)


andy-yx-chen:
Having a maxttl does not make sense to me, all you want to do is to prevent some bad connection from returning from pool. One simple way to solve the problem is to do a ping, the ping query could be configured, or nothing can prevent you default the ping query as "select 1 as alive", if the ping returns failure, then drop the connection from the pool, that would be better, cutting connections by a hard deadline is not a good idea

akleiman:
@andy-yx-chen It's not bad connections; we want to force the pool to recycle connections, even if they're good. Our use-case is: we have a query-heavy app and are running PG in an auto-scale group, behind a load-balancer. When the database CPU starts to go up, we spin up a new instance. Even setting an idle timeout doesn't help, because under heavy load the connections are never idle, so they're never disposed, and so the load balancer can't properly balance connections. So we want the pool to let connections go when they're older than a certain age, so it can attempt to reconnect.

This is something we had when we were using JDBC with HikariCP but lost in the move to jasync.

andy-yx-chen:
@akleiman sounds really fair then, I did not consider load balancer case just now

db-async-common/src/main/java/com/github/jasync/sql/db/ConnectionPoolConfiguration.kt
```diff
@@ -56,6 +57,7 @@ data class ConnectionPoolConfiguration @JvmOverloads constructor(
     val username: String = "dbuser",
     val password: String? = null,
     val maxActiveConnections: Int = 1,
+    val maxTtl: Long = -1,
```
oshai: I prefer calling it maxConnectionTtl. In addition, it is better to use nullable type with null as default to represent "no value". Please also add validation on allowed values (see below)

db-async-common/src/main/java/com/github/jasync/sql/db/pool/ActorBasedObjectPool.kt
```diff
@@ -496,6 +498,10 @@ private class ObjectPoolActor<T : PooledObject>(
         when (tried) {
             is Failure -> throw tried.exception
         }
+        val age = System.currentTimeMillis() - a.creationTime
+        if (configuration.maxTtl > 0 && age > configuration.maxTtl) {
+            throw ObjectAgedOutException(a, age, configuration.maxTtl)
```
oshai: I don't think we should throw an exception in case of ttl passed, just destroy the item on validation

db-async-common/src/main/java/com/github/jasync/sql/db/pool/ObjectAgedOutException.kt
```diff
@@ -0,0 +1,4 @@
+package com.github.jasync.sql.db.pool
+
+class ObjectAgedOutException(obj: PooledObject, age: Long, maxTtl: Long) :
```
oshai: This exception is not necessary in my opinion

db-async-common/src/main/java/com/github/jasync/sql/db/pool/ActorBasedObjectPool.kt
akleiman: Should validate() call dispose directly then?

db-async-common/src/main/java/com/github/jasync/sql/db/pool/ObjectAgedOutException.kt
akleiman: It exists only so I can hook into the code that calls validate, since they call dispose on exception.

db-async-common/src/main/java/com/github/jasync/sql/db/ConnectionPoolConfiguration.kt
akleiman: I'll call it `maxObjectTtl` in `PoolConfiguration` and `maxConnectionTtl` in the ConnectionPoolConfiguration/Builder, ok?

oshai: 👍

db-async-common/src/main/java/com/github/jasync/sql/db/pool/ActorBasedObjectPool.kt
akleiman: Calling `item.destroy()` directly, let me know if that's not what you had in mind.

oshai: Looking at the code again it seems that you did the right thing. The only comment then is to call the exception MaxTtlPassedException

db-async-common/src/main/java/com/github/jasync/sql/db/pool/ObjectAgedOutException.kt
oshai: see the comment above

db-async-common/src/main/java/com/github/jasync/sql/db/pool/ActorBasedObjectPool.kt
akleiman: :+1:

oshai:
Thanks again for the PR! I will do some more cleanups before release.
One thing that I want to test also is that it is not throwing an exception to the caller executing the query itself (which I suspect it is). If that is the case I will change the code, because the cleanup should not affect the one using the connection.

akleiman:
I ran some tests myself; if you set maxTTL very low (5-10 ms) then you'll see exceptions, since it'll age out almost as soon as you get the connection. If you raise it to something more reasonable (I went to 30s) then I didn't see exceptions.

oshai:
I am pretty sure you'll be able to see that on a busy pool in any configuration. I have some code changes ready, just fixing the tests now.

PR #99, opened 2019-03-04
initial r2dbc mysql support

oshai:


r2dbc-mysql/src/main/java/SimpleStatement.kt
```diff
@@ -0,0 +1,53 @@
...
+
+            val jasyncConnection = clientSupplier.get();
+
+            jasyncConnection.sendQuery(this.sql).handle { a, t: Throwable? ->
+
+                if (t == null) {
+
+                    val result = a.rows
+                    result.forEach {
+                        emitter.onNext(JaysncResult(result));
```
oshai: This line is not clear to me. why are we sending the result more than once?

oshai: @mp911de - if you can help with that it will be great, as well as general review. I didn't add transactions support yet.

build.gradle
```diff
@@ -15,12 +15,14 @@ buildscript {
 allprojects {
 
     group 'com.github.jasync-sql'
-    //version '0.8.39-netty4.0'
-    version '0.8.39'
+    version '0.0.0'
```
andy-yx-chen: why reset the version number

oshai: As you can see version number was not latest anyway. Wanted to prevent accidently overriding it.

r2dbc-mysql/src/main/java/JasyncClientConnection.kt
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    override fun createStatement(sql: String): Statement {
+        return SimpleStatement(this, sql)
+    }
+
+    override fun close(): Publisher<Void> {
+        return Flowable.create({ emitter ->
+            jasyncConnection.disconnect().thenApply {
+                emitter.onComplete()
```
mp911de: RxJava 2 is based on Java 6 so there's no utility to properly adapt a `CompletableFuture`. You might want to handle de error case as well (`emitter.onError(…)`).

r2dbc-mysql/src/main/java/SimpleStatement.kt
mp911de: A statement can contain two SQL operations (such as `SELECT * FROM foo; SELECT * FROM bar` or `INSERT INTO foo VALUES(…); UPDATE bar SET foo = 'bar';…`). Therefore, a database can return more than one result.

oshai: Got it. Our driver does not support multiple result set, so I will change that.

oshai:
@mp911de I added prepared statement support. you're welcome to review it and if it looks ok I will merge it and release a version, so it can be tested externally.

r2dbc-mysql/src/main/java/JasyncMetadata.kt
```diff
@@ -0,0 +1,65 @@
+package com.github.jasync.r2dbc.mysql
+
+import com.github.jasync.sql.db.ResultSet
+import io.r2dbc.spi.ColumnMetadata
+import io.r2dbc.spi.RowMetadata
+import java.util.*
+
+
+class JasyncMetadata(rows: ResultSet) : RowMetadata {
```
mp911de: Fyi, relates to r2dbc/r2dbc-spi#56

r2dbc-mysql/src/main/java/JasyncClientConnection.kt
```diff
@@ -0,0 +1,66 @@
...
+import io.r2dbc.spi.Connection
+import io.r2dbc.spi.IsolationLevel
+import io.r2dbc.spi.Statement
+import io.reactivex.BackpressureStrategy
+import io.reactivex.Flowable
+import org.reactivestreams.Publisher
+import java.util.function.Supplier
+import com.github.jasync.sql.db.Connection as JasyncConnection
+
+class JasyncClientConnection(private val jasyncConnection: JasyncConnection) : Connection, Supplier<JasyncConnection> {
```
mp911de: Do you have support for MySQL's PING? Adding this functionality here would be a preparation for r2dbc/r2dbc-spi#54.

r2dbc-mysql/src/main/java/JasyncRow.kt
```diff
@@ -0,0 +1,36 @@
...
+import io.r2dbc.spi.Row
+
+
+class JasyncRow(private val rowData: RowData) : Row {
+
+
+    override fun <T> get(identifier: Any, requestedType: Class<T>): T? {
+
+
+        if (identifier is String) {
```
mp911de: Initially dropping the requested type (`Class<T>`) is okay, but make sure to consider the requested type at some point in time as this is part of the SPI contract, otherwise, the calling side will run into `ClassCastException`.

r2dbc-mysql/src/main/java/JaysncResult.kt
```diff
@@ -0,0 +1,23 @@
...
+import com.github.jasync.sql.db.RowData
+import io.r2dbc.spi.Result
+import io.r2dbc.spi.RowMetadata
+import io.reactivex.Flowable
+import org.reactivestreams.Publisher
+import java.util.function.BiFunction
+
+
+class JaysncResult(private val resultSet: ResultSet) : Result {
+    private val metadata = JasyncMetadata(resultSet)
```
mp911de: 👍

r2dbc-mysql/src/main/java/SimpleStatement.kt
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    override fun bind(identifier: Any, value: Any): Statement {
+        TODO("not implemented")
+    }
+
+    override fun bind(index: Int, value: Any): Statement {
+        isPrepared = true
+        when {
+            index > params.size -> throw IllegalArgumentException("can't skip params when binding")
```
mp911de: I'd propose to keep bindings in a `Map`-like structure. The SPI does not require an ascending binding order. Callers may bind arguments indexes that do not follow an ordering (`bind(0, …).bind(17, …).bind(1, …).bind(2).bind(16,…)`)

r2dbc-mysql/src/main/java/SimpleStatement.kt
```diff
@@ -0,0 +1,67 @@
...
+            index == params.size -> params.add(value)
+            else -> params[index] = value
+        }
+        return this
+    }
+
+    override fun bindNull(identifier: Any, type: Class<*>): Statement {
```
mp911de: What is MySQL's native bind maker?

r2dbc-mysql/src/main/java/JaysncResult.kt
```diff
@@ -0,0 +1,23 @@
...
+
+    override fun getRowsUpdated(): Publisher<Int> {
+        return Flowable.just(resultSet.size)
+    }
+
+    override fun <T> map(mappingFunction: BiFunction<io.r2dbc.spi.Row, RowMetadata, out T>): Publisher<T> {
+        return Flowable.fromIterable<RowData>(resultSet)
```
mp911de: Do you read all rows at once and return afterwards a `ResultSet`? Or is `ResultSet` fetching data when it reaches the end of e.g. a cursor page?

r2dbc-mysql/src/main/java/JasyncClientConnection.kt
oshai: the connection itself don't have ping, but there is a connection factory that knows how to test the connection: https://github.com/jasync-sql/jasync-sql/blob/dde3f6afeaa325189ed0467bfd1342bdce9adde9/db-async-common/src/main/java/com/github/jasync/sql/db/pool/ConnectionFactory.kt#L79

r2dbc-mysql/src/main/java/JasyncRow.kt
oshai: I'm not sure I understand the spi. should the returned value be of type `requestedType`?

r2dbc-mysql/src/main/java/SimpleStatement.kt
oshai: and to enforce it on the execution itself?

r2dbc-mysql/src/main/java/SimpleStatement.kt
oshai: can you explain what you mean?

r2dbc-mysql/src/main/java/JaysncResult.kt
oshai: there is no inherent paging in the driver at the moment and all results are fetched before the future completes

r2dbc-mysql/src/main/java/JasyncClientConnection.kt
mp911de: The idea behind is to check for e.g. pooling implementations whether the connection works without actually executing SQL.

r2dbc-mysql/src/main/java/JasyncRow.kt
mp911de: Exactly. Each database type maps to a specific Java type (e.g. `TINYINT` to `Byte`, `SMALLINT` to `Short`). You should be able to retrieve a `TINYINT` as `Integer.class` or as `Long.class`. The caller requests the type `Integer.class` and expects the `get(Any, Class<T>)` method to return that specific type.

This conversion makes only sense for convertible types (`Numbers`, Dates (`LocalDateTime` to `LocalTime` or `LocalDate`) and converting values to `String`).

For a start, if you don't want to implement an advanced Codec mechanism, it would make sense to consider the mentioned three conversions and if the value is not `null`, to `requestedType.cast(…)` the value.

In contrast,  `get(Any)` returns the native type (i.e. what happens right now, that the driver determines the most appropriate type).

r2dbc-mysql/src/main/java/SimpleStatement.kt
mp911de: Yeah, that makes sense. Give freedom during the binding time and require proper binding on `execute()` and on `add()` (Note: Add records the current binding and prepares the statement to add another set of bindings so that `execute()` calls the statement for each binding, sort of batching). Implementation-wise you would iterate over bindings and issue a new query via something like `Flowable.fromIterable(bindings).flatMap(it -> executePreparedQuery(it))`.

r2dbc-mysql/src/main/java/SimpleStatement.kt
mp911de: Sure. Postgres uses dollar-notation (`$1`, `$2`) to indicate parameter placeholders (`SELECT name FROM person WHERE firstname = $1`), SQL server uses at-prefixed named parameters (`@foo`, `@P0`, `@P1`, `SELECT name FROM person WHERE firstname = @P0`).

r2dbc-mysql/src/main/java/JaysncResult.kt
mp911de: That's good to know and that is a good behavior already (could be an optimization in the future).

Some drivers (like Cassandra) have a `ResultSet` and as soon as the iterator reaches the end of a page, the iterator it goes into a blocking mode to fetch the next page. That's a nasty surprise because it's not obviously visible, and that was the background of my question.

r2dbc-mysql/src/main/java/JasyncMetadata.kt
oshai: I am not sure if that is what you mean but in the driver we have col names: https://github.com/jasync-sql/jasync-sql/blob/dde3f6afeaa325189ed0467bfd1342bdce9adde9/db-async-common/src/main/java/com/github/jasync/sql/db/ResultSet.kt#L19

r2dbc-mysql/src/main/java/JasyncClientConnection.kt
oshai: no, there is no ping support, but there is a 'validate' check that don't do any network action: https://github.com/jasync-sql/jasync-sql/blob/dde3f6afeaa325189ed0467bfd1342bdce9adde9/db-async-common/src/main/java/com/github/jasync/sql/db/pool/ConnectionFactory.kt#L49

r2dbc-mysql/src/main/java/JasyncRow.kt
oshai: as you mentioned the driver has "builtin" types right now depending on the db type. I am not what you mean to do for a start? just a cast? didn't understand the different cases you mentioned.

r2dbc-mysql/src/main/java/SimpleStatement.kt
oshai: I added it

r2dbc-mysql/src/main/java/SimpleStatement.kt
oshai: in mysql it is just `?` so there is no indexing as far as I know.

r2dbc-mysql/src/main/java/JasyncMetadata.kt
mp911de: We will add the column names shortcut as additional SPI method, so this is just a heads-up.

r2dbc-mysql/src/main/java/JasyncRow.kt
mp911de: The method should do at least a cast. 

I would also suggest adding the following conversion rules:
* If a value is a `Number` and the requested type is a number, too, then convert the value to `T`  using `Number.intValue()`, `Number.longValue()` methods.
* If the requested type is `String`, then call `toString()` on the value.
* If a value is `LocalDate`, `LocalDateTime`, or `LocalTime` and the requested type is one of these date types, then convert the value into the requested value using the provided methods on `LocalDate` etc (e.g. `LocalDateTime.atStartOfDay()`).

Does this make sense?

r2dbc-mysql/src/main/java/SimpleStatement.kt
mp911de: Okay, I see. Then the only thing to do here is supporting indexes. If the identifier is a `String`, then the method would parse the `String` into an `Int`.

oshai: Added it.

r2dbc-mysql/build.gradle
```diff
@@ -0,0 +1,26 @@
+
+dependencies {
+    compile project(':mysql-async')
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
+    compile "io.r2dbc:r2dbc-spi:1.0.0.BUILD-SNAPSHOT"
+    implementation "io.reactivex.rxjava2:rxjava:2.2.7"
```
mirromutth: Is it compatible or something else? Otherwise using `reactor-core` may be a better choice because of more powerful and concise API.

r2dbc-mysql/src/main/java/JasyncRow.kt
oshai: yes, I added it, please review again to see we are aligned. Note that for DateTime the lib is using joda so I had to do an extra conversion.

r2dbc-mysql/build.gradle
oshai: I don't have a strong opinion about it, just followed what @mp911de did in this gist: https://gist.github.com/mp911de/9ea13939e8fd9a6b4ef138419f085715

mp911de: I used RxJava 2 in the vert.x example as vert.x had no dependency on Project Reactor. 
My choice for a Java 8 project would be Project Reactor, because it properly supports `CompletableFuture` (Java 8 types) and it has a `Context` that can be used to bind a connection to, if someone wants to properly manage a transaction.

r2dbc-mysql/src/main/java/JasyncRow.kt
mp911de: Looks pretty decent now. I wasn't aware that you build on top of Joda. For R2DBC we want to stay within the Java 8 type realm for primitives. The native `get(Any)` method should ideally return Java 8 types for dates. Retrieving a Joda type would be still possible via `get(Any, Class<T>)` when requesting the Joda `LocalDateTime`.

oshai: I will change that. we have joda as a legacy from the original driver.

r2dbc-mysql/build.gradle
oshai: I changed it.

mp911de:
Did a first integration test pass with Spring Data R2DBC and things look quite for simple queries using MySQL 5.6.43. The latest MySQL version fails to connect with `UnsupportedAuthenticationMethodException: Unknown authentication method -> 'caching_sha2_password'`.

```
MySQLConnectionFactory mycf = new MySQLConnectionFactory(
		new Configuration("root", "localhost", 3306, "my-secret-pw", "mysql"));

JasyncConnectionFactory cf = new JasyncConnectionFactory(mycf);

DatabaseClient client = DatabaseClient.builder().connectionFactory(cf)
		.dataAccessStrategy(new DefaultReactiveDataAccessStrategy(PostgresDialect.INSTANCE)).build();

client.execute()
		.sql("SELECT * FROM help_topic")
		.map((row, rowMetadata) -> row.get("example"))
		.all()
		.doOnNext(it -> {
			System.out.println("Record: " + it);
		})
		.as(StepVerifier::create)
		.expectNextCount(585)
		.verifyComplete();
```

For the `JasyncConnectionFactory` (if it will remain in a MySQL related package), it would be nicer to configure it with just `Configuration`, but that's pure cosmetics to me.

oshai:
caching_sha2_password is default auth since 8.0.4, but it can be changed on the db server: https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_default_authentication_plugin

Regarding configuration, let's not change it now but think of a consistent api later.

So I will merge this PR and let you know what it's initial released version.

PR #101, opened 2019-03-08
Avoid buf copy and fix buf leak

jilen:


mysql-async/src/main/java/com/github/jasync/sql/db/mysql/codec/MySQLConnectionHandler.kt
```diff
@@ -164,7 +164,9 @@ class MySQLConnectionHandler(
                     }
                     ServerMessage.BinaryRow -> {
                         val m = message as BinaryRowMessage
-                        this.currentQuery!!.addRow(this.binaryRowDecoder.decode(m.buffer, this.currentColumns))
+                        val decodedRow = this.binaryRowDecoder.decode(m.buffer, this.currentColumns)
+                        m.buffer.release()
```
andy-yx-chen: my understanding is you don't need to release the buffer from BytesToMessageDecoder

mysql-async/src/main/java/com/github/jasync/sql/db/mysql/codec/MySQLFrameDecoder.kt
```diff
@@ -222,7 +222,8 @@ class MySQLFrameDecoder(val charset: Charset, private val connectionId: String)
 
         return if (this.totalColumns == this.processedColumns) {
             if (this.isPreparedStatementExecute) {
-                val row = slice.readBytes(slice.readableBytes())
+                val row = slice.retainedSlice()
+                slice.readerIndex(slice.readerIndex() + slice.readableBytes())
```
andy-yx-chen: one concern is the buffer is attached to BufferNotFullyConsumedException in BinaryRowDecoder, which may become unavailable during the exception dispatching.

The original code may leak iff buffer that returned from slice.readBytes must be explicitly release, but looks like API doc does not mention

jilen: If not retain, the buf will be released while thie handler finished, and cannot be accessed in another handler.

jilen: @andy-yx-chen  Replaced with `readRetainedSlice`.

andy-yx-chen: @jilen you are right, please help to check if there are other places that have the same issue here

andy-yx-chen: @oshai , we need you to take a look at this one, good finding actually

oshai:
I will review it as soon as possible. I just wonder why this leak wasn't noticed until now. Will also compare it to mauricio.

andy-yx-chen:
> I will review it as soon as possible. I just wonder why this leak wasn't noticed until now. Will also compare it to mauricio.

thanks, I think the fix is valid, I also read the source code for netty, the readBytes will return a buffer from MessageDecoder's ctx.allocator, which is pooled, and need to be explicitly release

oshai:
@jilen Thanks for raising the issue!
I want to be sure that I understand the change: `readBytes` is not efficient because it is copying the buffer (BTW I found 3 places like that). Instead, you would like to use `readRetainedSlice` which is a view on the same buffer with it's own ref count thus more efficient.

In addition, the copied buffer was probably leaking. I tried using `-Dio.netty.leakDetection.level=paranoid` but there wasn't any message. On the other hand checking `m.buffer.refCnt()` after a while was == 1. If you can bring some more details or ideas how to make sure we can obeserve the leak it will be great.

What I did find is the following:
* `DataRowMessage` was leaking in the past but it was fixed: https://github.com/mauricio/postgresql-async/pull/181
* `ResultSetRowMessage.buffer` is probably also leaking. 

Would be glad to get more input before merging it.

jilen:
@oshai  Yes, those bufs could be replaced with `readRetainedSlice`, and be released later.  And netty's leak detection relies on gc. So it might not always reliable. For me, sometimes it will report leak, while others it will not.

andy-yx-chen:
again, whether detector reports the leak or not, the fact is readBytes in MessageDecoder will cause leak, as it will allocate a buffer in handler context.alloc, which needs to be release explicitly as the document says

oshai:
@jilen Thanks again for the fix!

PR #104, opened 2019-03-20
R2dbc fixes

mp911de:
This PR contains a couple of changes that are independent of each other but are required to enable missing functionality through the R2DBC SPI:

* Add implementation for transactional methods 
* Fix row type mapping
* Defer calls that involve CompletableFuture adaption
* Report affected rows count


oshai:
Thanks for the fixes!

PR #107, opened 2019-03-21
A mini PR, to simplify lazy loading of `execute`

mirromutth:
This may reduce the creation of temporary objects (which of course has little effect), but this seems to be more in line with the *philosophy* of the Reactor.


oshai:
Thanks!

PR #108, opened 2019-03-21
Support `returnGeneratedValues` for `INSERT` statement

mirromutth:
MySQL only support `SELECT LAST_INSERT_ID()` when last query has changed any values of `AUTO_INCREMENT`

### DIFF
--- a/r2dbc-mysql/src/main/java/SimpleStatement.kt
+++ b/r2dbc-mysql/src/main/java/SimpleStatement.kt
@@ -1,5 +1,6 @@
 package com.github.jasync.r2dbc.mysql
 
+import com.github.jasync.sql.db.QueryResult
 import io.r2dbc.spi.Result
 import io.r2dbc.spi.Statement
 import org.reactivestreams.Publisher
@@ -12,6 +13,7 @@ class SimpleStatement(private val clientSupplier: Supplier<JasyncConnection>, pr
 
     private var isPrepared = false
     private var params: MutableMap<Int, Any?> = mutableMapOf()
+    private var selectLastInsertId: Boolean = false
 
     override fun add(): Statement {
         TODO("not implemented")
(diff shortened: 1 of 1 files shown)


oshai:
@mp911de - can you please review that this is the expected api?

r2dbc-mysql/src/main/java/SimpleStatement.kt
```diff
@@ -45,9 +47,40 @@ class SimpleStatement(private val clientSupplier: Supplier<JasyncConnection>, pr
         return this
     }
 
+    override fun returnGeneratedValues(vararg columns: String): Statement {
+        if (columns.any { !it.equals("LAST_INSERT_ID", true) }) {
```
mp911de: The intention of `columns` is to specify a desired column name for generated keys. So the first column name should be used as an alias (`SELECT LAST_INSERT_ID() as $columns[0]`). See [SQL Server implementation](https://github.com/r2dbc/r2dbc-mssql/blob/master/src/main/java/io/r2dbc/mssql/GeneratedValues.java#L112) for further details.

r2dbc-mysql/src/main/java/SimpleStatement.kt
```diff
@@ -45,9 +47,40 @@ class SimpleStatement(private val clientSupplier: Supplier<JasyncConnection>, pr
...
+
+        return this
+    }
+
     override fun execute(): Publisher<out Result> {
         val jasyncConnection = clientSupplier.get()
+        val queried = queryExecute(jasyncConnection)
+
+        if (selectLastInsertId) {
+            return queried.flatMap { result ->
```
mp911de: Is it possible to combine `SELECT LAST_INSERT_ID()` with the insert statement? E.g. appending `SELECT LAST_INSERT_ID()` to the insert query so we get just a single server roundtrip?

r2dbc-mysql/src/main/java/SimpleStatement.kt
```diff
@@ -62,7 +95,7 @@ class SimpleStatement(private val clientSupplier: Supplier<JasyncConnection>, pr
             } else {
                 jasyncConnection.sendQuery(this.sql)
             }
-            r.toMono().map { JaysncResult(it.rows, it.rowsAffected) }
+            r.toMono()
```
oshai: I think the way to go (to save the additional roundtrip) is to cast the result here to `MySQLQueryResult` and use the `lastInsertId` property.

mp911de:
I'll come back to this PR later today.

r2dbc-mysql/src/main/java/SimpleStatement.kt
mp911de: There are two checks for `selectLastInsertId`, a single check is sufficient. I also slightly changed the code:

```kotlin
         if (selectLastInsertId) {
            return queried.flatMap { result ->
                    jasyncConnection.sendQuery("SELECT LAST_INSERT_ID() AS $generatedKeyName")
                        .toMono()
                        .map { JaysncResult(it.rows, result.rowsAffected) }
            }
        }
```

r2dbc-mysql/src/main/java/SimpleStatement.kt
mp911de: That can be simplified to:

```kotlin
        if (columns.size == 1) {
            generatedKeyName = columns[0]
        }

        if (columns.size > 1) {
            throw IllegalArgumentException("MySQL only supports a single generated value")
        }

        selectLastInsertId = true

        return this
```

r2dbc-mysql/src/main/java/SimpleStatement.kt
```diff
@@ -12,6 +13,7 @@ class SimpleStatement(private val clientSupplier: Supplier<JasyncConnection>, pr
 
     private var isPrepared = false
     private var params: MutableMap<Int, Any?> = mutableMapOf()
+    private var selectLastInsertId: Boolean = false
```
mp911de: We should also add `private var generatedKeyName: String = "LAST_INSERT_ID"` here.

oshai:
@mp911de what about using the value returned by the insert statement as I suggested?

mp911de:
From a client perspective, it's desirable to have some control over the generated key name. Otherwise, a client has always to inspect metadata for the results. So the idea behind the columns is to give clients an ability to indicate the desired column name. Drivers can pick it up and apply the hint.

It will work also without a name hint, but it is nicer with a name hint.

oshai:
Ok. Let me know when it is ready to be merged.

mp911de:
Ping @mirromutth

oshai:
@mp911de - I added the changes you suggested, let me know if it looks good and I will merge it.

r2dbc-mysql/src/main/java/SimpleStatement.kt
oshai: added below

mp911de:
Looks good now!

oshai:
Thanks!

Issue #110, opened 2019-03-25
About implementation of r2dbc `Batch`


mirromutth:
At the moment, `jasync` may not support executing a single query containing multiple statements. (to exchange multi-results)

@mp911de @oshai Could just support `Batch` that is a "fake" batch logic, execute multi-statements one-by-one? Like this: https://github.com/jasync-sql/jasync-sql/pull/109

mp911de:
The SPI specifies mainly the intent to invoke certain things. `Batch` takes multiple simple SQL queries without binding and executes these. If MySQL or jasync do not support batch frames, then It should be fine to issue multiple regular queries (e.g. `List<String> sqlStatements` exected as`Flux.from(sqlStatements).flatMap(sql -> Mono.fromFuture(jasync.sendQuery(sql)).map(…))`)

The same approach should work with `Statement.add()` and multiple bindings.

oshai:
I will merge #109 

Do you think there is a simple way to have the r2dbc tests in the module so we will make sure nothing gets broken?

mp911de:
I would suggest in any case to add MockK tests, R2DBC Client has embedded test suites (see https://github.com/r2dbc/r2dbc-client/blob/master/src/test/java/io/r2dbc/client/MssqlExample.java for SQL Server). You can add these tests either to your project or submit a PR to R2DBC Client.

mirromutth:
> The same approach should work with `Statement.add()` and multiple bindings.

@oshai I will submit a new PR to support `Statement.add()` later.

oshai:
I added an example here: https://github.com/r2dbc/r2dbc-client/pull/46

Closing this issue.

PR #111, opened 2019-03-26
Support `Statement.add()` for multiple bindings

mirromutth:
See issue https://github.com/jasync-sql/jasync-sql/issues/110

### DIFF
--- a/r2dbc-mysql/src/main/java/Bindings.kt
+++ b/r2dbc-mysql/src/main/java/Bindings.kt
@@ -1,0 +1,29 @@
+package com.github.jasync.r2dbc.mysql
+
+/**
+ * Statement parameters bindings.
+ */
+internal class Bindings {
+
+    private val bindings: MutableList<Map<Int, Any?>> = mutableListOf()
+
+    private var current: MutableMap<Int, Any?>? = null
+
+    fun all(): List<Map<Int, Any?>> = bindings
+
+    fun current(): MutableMap<Int, Any?> {
+        var current = this.current
+
+        if (current == null) {
+            current = mutableMapOf()
+            this.current = current
+            this.bindings.add(current)
+        }
+
+        return current
+    }
+
+    fun done() {
+        current = null
+    }
+}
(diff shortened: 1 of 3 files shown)


r2dbc-mysql/src/main/java/ExtendedStatement.kt
```diff
@@ -5,15 +5,17 @@ import io.r2dbc.spi.Result
 import io.r2dbc.spi.Statement
 import org.reactivestreams.Publisher
 import reactor.core.publisher.Mono
+import reactor.core.publisher.toFlux
 import reactor.core.publisher.toMono
-import java.util.concurrent.CompletableFuture
 import java.util.function.Supplier
 import com.github.jasync.sql.db.Connection as JasyncConnection
 
-class SimpleStatement(private val clientSupplier: Supplier<JasyncConnection>, private val sql: String) : Statement {
```
oshai: I think calling it StatementImpl might be more appropriate (or JasyncStatement)

r2dbc-mysql/src/main/java/Bindings.kt
```diff
@@ -0,0 +1,29 @@
+package com.github.jasync.r2dbc.mysql
+
+/**
+ * Statement parameters bindings.
+ */
+internal class Bindings {
+
+    private val bindings: MutableList<Map<Int, Any?>> = mutableListOf()
```
oshai: how about calling it paramBindings?

r2dbc-mysql/src/main/java/ExtendedStatement.kt
```diff
@@ -57,40 +63,35 @@ class SimpleStatement(private val clientSupplier: Supplier<JasyncConnection>, pr
...
-            if (selectLastInsertId) {
-                queried.flatMap { result ->
-                    connection.sendQuery("SELECT LAST_INSERT_ID() AS $generatedKeyName")
-                        .toMono()
-                }
+        return Mono.fromSupplier(clientSupplier).flatMapMany { connection ->
+            if (isPrepared) {
+                val allParams = bindings.all().asSequence().mapIndexed { i, binding ->
+                    (0 until binding.size).map {
+                        if (it in binding) {
```
oshai: nice

oshai:
@mp911de - can you please review that as well?

r2dbc-mysql/src/main/java/Bindings.kt
mp911de: `bindings` lines nicely up with what we have in the other drivers as well.

r2dbc-mysql/src/main/java/ExtendedStatement.kt
mp911de: 👍 for `JasyncStatement`

PR #112, opened 2019-03-26
Add MysqlConnectionFactoryProvider

oshai:


mp911de:
The change isn't complete yet. You need to provide a service loader registration (see https://github.com/r2dbc/r2dbc-postgresql/blob/master/src/main/resources/META-INF/services/io.r2dbc.spi.ConnectionFactoryProvider) for `MysqlConnectionFactoryProvider`. Once `META-INF/services/io.r2dbc.spi.ConnectionFactoryProvider` is in place, then the examples for `r2dbc-client` will get past the error.

PR #115, opened 2019-04-04
change the default utf8 encoding

oshai:
according to this post https://mathiasbynens.be/notes/mysql-utf8mb4
the default utf8 in mysql do not exactly match utf8 standard (3 vs 4 bytes)

Since it maps to the same charset in Java
It sounds like the reasonable thing to do is to change the used protocol
from 83 to 224 (utf8mb4_unicode_ci)
I added a flag to allow overriding it `-DjasyncMysqlUTF8Collation=83`
To put in back in the old legacy behaviour
for example if using mysql server < 5.5.3.

Added also tests for this.

This PR, is a fix for the issue raised here: https://github.com/vert-x3/vertx-mysql-postgresql-client/issues/149

### DIFF
--- a/mysql-async/src/main/java/com/github/jasync/sql/db/mysql/util/CharsetMapper.kt
+++ b/mysql-async/src/main/java/com/github/jasync/sql/db/mysql/util/CharsetMapper.kt
@@ -11,7 +11,7 @@ class CharsetMapper(charsetsToIntComplement: Map<Charset, Int> = emptyMap()) {
         const val Binary = 63
 
         val DefaultCharsetsByCharset = mapOf(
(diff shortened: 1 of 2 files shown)


oshai:
@swordfeng - would like to get your opinion about this change. Thanks!

swordfeng:
@oshai Thank you! It looks good to me. Maybe we could add that flag to the document in case someone is using a version of MySQL released 9 years ago.

oshai:
released in 0.9.50 and add a wiki here: https://github.com/jasync-sql/jasync-sql/wiki/Supported-Encoding

PR #125, opened 2019-06-04
add logging interceptor

oshai:


oshai:
see docs here: https://github.com/jasync-sql/jasync-sql/wiki/Interceptors#logging-support

Issue #128, opened 2019-06-22
[R2dbc] Consider property `lastInsertId` of `OkMessage` to implement `returnGeneratedValues`


mirromutth:
It is a same issue from https://github.com/mirromutth/r2dbc-mysql/issues/21.

`lastInsertId` is a property in `OkMessage` (server-side). After the client sends `INSERT` statement, it will be a generated value same as the value of `SELECT LAST_INSERT_ID()` result.

If last statement is not `INSERT`, this property will be 0 same as the value of `SELECT LAST_INSERT_ID()` result.

So, can just use `OkMessage.lastInsertId` in statement result to consider `returnGeneratedValues` implemented by the driver level, rather than `SELECT LAST_INSERT_ID()` which is the application level. This will improve the performance and *atomic* operation.

This is also more friendly to prepared statements because the prepared statements do **NOT** support batch statements which not like simple statements.

oshai:
@mirromutth - any plan to contribute that addition?

mirromutth:
@oshai I am writing the code of relevant with `returnGeneratedValues`, but the exact time is not clear (because of my work or other things), maybe next weekend.

oshai:
Thanks!

mirromutth:
@oshai I might want to set `com.github.jasync.sql.db.mysql.MySQLQueryResult` to `public` for get `MySQLQueryResult.lastInsertId`, or do you have a better idea?

PR #130, opened 2019-06-24
Support `returnGeneratedValues` by `lastInsertId`

mirromutth:
This PR is for issue #128

In addition, MySQL does not support negative integer or `0` for `auto_increment`, so `lastInsertId` always be a positive integer.

### DIFF
--- a/r2dbc-mysql/src/main/java/JasyncInsertFakeMetadata.kt
+++ b/r2dbc-mysql/src/main/java/JasyncInsertFakeMetadata.kt
@@ -1,0 +1,46 @@
+package com.github.jasync.r2dbc.mysql
+
+import io.r2dbc.spi.ColumnMetadata
+import io.r2dbc.spi.RowMetadata
+
+/**
+ * An implementation of [RowMetadata] for support [JasyncStatement.returnGeneratedValues].
+ *
+ * It is also an implementation of [ColumnMetadata] for reduce redundant objects.
+ *
+ * @see JasyncInsertFakeRow
+ */
+internal class JasyncInsertFakeMetadata(private val generatedKeyName: String) : RowMetadata, ColumnMetadata {
+
+    override fun getColumnMetadatas(): Iterable<ColumnMetadata> {
+        return listOf(this)
+    }
+
+    override fun getColumnMetadata(identifier: Any): ColumnMetadata {
+        return when (identifier) {
+            is Int -> {
+                when {
+                    identifier > 0 -> throw ArrayIndexOutOfBoundsException("Column index $identifier is larger than the number of columns 1")
+                    identifier < 0 -> throw ArrayIndexOutOfBoundsException("Column index $identifier is negative")
+                    else -> this
+                }
+            }
+            is String -> {
+                if (identifier != generatedKeyName) {
+                    throw NoSuchElementException("Column name '$identifier' does not exist in column names {$generatedKeyName}")
+                }
+
+                this
+            }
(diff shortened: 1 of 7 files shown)


oshai:
@mirromutth LGTM. @mp911de any comments from your side?

r2dbc-mysql/src/main/java/JasyncInsertFakeMetadata.kt
```diff
@@ -0,0 +1,46 @@
...
+import io.r2dbc.spi.RowMetadata
+
+/**
+ * An implementation of [RowMetadata] for support [JasyncStatement.returnGeneratedValues].
+ *
+ * It is also an implementation of [ColumnMetadata] for reduce redundant objects.
+ *
+ * @see JasyncInsertFakeRow
+ */
+internal class JasyncInsertFakeMetadata(private val generatedKeyName: String) : RowMetadata, ColumnMetadata {
```
mp911de: How about calling it `Synthetic` or `GeneratedKeys` instead of `Fake`?

r2dbc-mysql/src/main/java/JasyncInsertFakeRow.kt
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * A fake row considers last inserted ID for support [JasyncStatement.returnGeneratedValues].
+ *
+ * @see JasyncResult
+ * @see JasyncStatement
+ * @see JasyncInsertFakeMetadata
+ * @see io.r2dbc.spi.Statement
+ */
+internal class JasyncInsertFakeRow(private val generatedKeyName: String, private val lastInsertId: Long) : Row {
```
mp911de: Same naming issue here.

r2dbc-mysql/src/main/java/JasyncInsertFakeMetadata.kt
```diff
@@ -0,0 +1,46 @@
...
+                    throw NoSuchElementException("Column name '$identifier' does not exist in column names {$generatedKeyName}")
+                }
+
+                this
+            }
+            else -> throw IllegalArgumentException("Identifier '$identifier' is not a valid identifier. Should either be an Integer index or a String column name.")
+        }
+    }
+
+    override fun getColumnNames(): Collection<String> {
```
mp911de: `getColumnNames` should be case-insensitive when matching against column names (see interface Javadoc).

oshai:
I agree that syntetic sounds more appropriate.

r2dbc-mysql/src/main/java/JasyncInsertFakeMetadata.kt
mirromutth: Now naming it `JasyncInsertSyntheticMetadata`.

r2dbc-mysql/src/main/java/JasyncInsertFakeRow.kt
mirromutth: Naming it `JasyncInsertSyntheticRow` now.

r2dbc-mysql/src/main/java/JasyncInsertFakeMetadata.kt
mirromutth: This is my fault and I shall check also in my project, it is using `TreeSet(String.CASE_INSENSITIVE)` for `jasync-sql`.

oshai:
@mirromutth - is it ready to be merged?

mirromutth:
> @mirromutth - is it ready to be merged?

I think it is ready.

PR #132, opened 2019-07-01
Add scalikejdbc-async to the list of users on README

seratch:
Thanks for making a great successor of postgresql/mysql-async. We've switched to this library. Could you have us as one of the users on README?

https://github.com/scalikejdbc/scalikejdbc-async/pull/74


oshai:
Thanks!

Issue #135, opened 2019-07-10
java.util.NoSuchElementException: Key LAST_INSERT_ID is missing in the map.


set4812:
Hi,
After went this tutorial https://medium.com/w-logs/jdbc-for-spring-webflux-spring-data-r2dbc-99690208cfeb and upgraded packages, project run without problem. Rows inserting, updating and quering without problems. I was tried check how batch inserting will be work and I getting error after insert a few records into DB
`java.util.NoSuchElementException: Key LAST_INSERT_ID is missing in the map.`. When i downgrade version into 1.0.1 there is no error, bulk works without problems. Code for error replication
https://github.com/set4812/R2DBC-error

I think this PR broken this
https://github.com/jasync-sql/jasync-sql/pull/130

mirromutth:
I'm so sorry for your trouble, this is my stupid mistake of missing a boolean inversion.

I have submitted a PR to fix the bug.

set4812:
Looks great, now maintainer need approve na merge.

oshai:
Fix released in 1.0.4

set4812:
@mirromutth after fix i have
```
java.lang.IllegalStateException: unmatched requested type Object
	at com.github.jasync.r2dbc.mysql.JasyncInsertSyntheticRow.getValue(JasyncInsertSyntheticRow.kt:59) ~[jasync-r2dbc-mysql-1.0.4.jar:na]
	at com.github.jasync.r2dbc.mysql.JasyncInsertSyntheticRow.get(JasyncInsertSyntheticRow.kt:31) ~[jasync-r2dbc-mysql-1.0.4.jar:na]
	at io.r2dbc.spi.Row.get(Row.java:63) ~[r2dbc-spi-0.8.0.M8.jar:na]
	at org.springframework.data.r2dbc.convert.MappingR2dbcConverter.potentiallySetId(MappingR2dbcConverter.java:409) ~[spring-data-r2dbc-1.0.0.BUILD-20190710.100158-150.jar:1.0.0.BUILD-SNAPSHOT]
```
Maybe error trace will help. On 1.0.1 its working nice with spring data r2bc.

mirromutth:
@set4812 I create a new issue #137 and PR #138 for the problem which is "unmatched requested type Object".

I think this is convenient for us to track these two different issues.

set4812:
I closing this as fixed reported error

PR #136, opened 2019-07-11
Correct condition for synthetic row

mirromutth:
This is my stupid mistake of missing a boolean inversion.

Correct the condition come from #130 .

See issue #135 .


oshai:
Thanks for the fix, I think I will create tests for the r2dbc module.

Issue #131, opened 2019-06-30
remove joda-time dependency


xuwei-k:
I think we should use `java.time.*` classes instead of joda-time

oshai:
I think it is a good idea. On the other hand, I am afraid backwards-incompatibility will prevent users from upgrading and might be risky for them. So if we can come up with a smooth migration path it can be helpful. Another alternative is to support both jdk/joda somehow without having to maintain two branches.

xp-vit:
sad but true: joda-time need to be dropped. And I think it makes sense to do it as soon as possible, because it is going to be breaking change and number of affected users should be as small as possible. Options on how to implement this change are:
1. Radical(preferable):
 - major release (or at least version 1.1) and just remove all joda-time references
2. Soft:
 - deprecate all public API which currently returns joda-time and add new methods which will use java.time.*
 - announce that deprecated API will be dropped in future major release
 - drop deprecated api together with joda-time dependency in next major release

oshai:
The second alternative is really not feasible for that lib. joda time can be returned from rows in result set so it is hard to replace it there and can be given as a parameter to a prepared statement.
Another idea is to have some kind of "feature flag" that based on that we will decide which one to use.
Then users can pick which one they want, and in case they do not want joda they will have to exclude it from the dependency in gradle/maven.

xp-vit:
no... excluding dependency is dirty and complicated. Only 1% of users will do it I believe... Only those who care about classpath size. I would try to go with more radical way then, since library is really young and there are not a lot users yet.

oshai:
The library is almost 1 year old and it is a fork of [mauricio lib](https://github.com/mauricio/postgresql-async) that existed since 2012. We want to allow easy migration from the original lib if possible. I guess the original one has many users.
So from my perspective, it is better to allow an easy and safe migration path if possible.

The question is what will be the benefit of such a move? From my perspective, it can be avoiding classpath collisions and moving to the standard api. Other than that, I don't see great benefits from that change. Are there?

So although I think it is a good idea (as I mentioned above) I am pretty reluctant to go with the radical approach at the moment.

I am still open to other ideas about how to do the migration.

mirromutth:
How about change dependency from a requirement to an optional one? Like `compileOnly('...joda-time...')`, that means if user dependent on joda, than enable joda, otherwise disable for default.

Than, user can just upgrade jasync-sql version and add a dependency of joda rather than change their code.

oshai:
[...]

PR #157, opened 2019-09-30
[WIP] Removes JodaTime

bakatz:
I don't want a dependency on JodaTime for my projects, so I decided to start fixing #131 

TODO:
[x] Common library
[x] MySQL
[x] Fix TimestampEncoder
[ ] Fix MySQL tests
[] Postgres
[] Fix Postgres tests

Looking for comments on this before I finish the whole thing (ideally from someone who has in depth knowledge of `java.time`) - I'm not 100% sure I've converted the calls successfully in some cases. Mostly followed this as a guide: https://blog.joda.org/2014/11/converting-from-joda-time-to-javatime.html

Implementation notes:
~TimestampEncoder is definitely broken, probably need to use ZonedDateTime instead of LocalDateTime etc. Will take a look at it later and resolve.~ Even though TimestampEncoder assumes no time zone, it formats times using a 'Z' suffix to represent the UTC timezone. As such, I had to make it use `ZonedDateTime` otherwise a formatting exception would occur at runtime.

Cheers

### DIFF
--- a/db-async-common/src/main/java/com/github/jasync/sql/db/column/InetAddressEncoderDecoder.kt
+++ b/db-async-common/src/main/java/com/github/jasync/sql/db/column/InetAddressEncoderDecoder.kt
@@ -1,18 +1,10 @@
 package com.github.jasync.sql.db.column
 
-import sun.net.util.IPAddressUtil.textToNumericFormatV4
-import sun.net.util.IPAddressUtil.textToNumericFormatV6
 import java.net.InetAddress
 
 object InetAddressEncoderDecoder : ColumnEncoderDecoder {
 
-    override fun decode(value: String): Any {
-        return if (value.contains(':')) {
-            InetAddress.getByAddress(textToNumericFormatV6(value))
-        } else {
-            InetAddress.getByAddress(textToNumericFormatV4(value))
-        }
-    }
+    override fun decode(value: String): Any = InetAddress.getByName(value)
 
     override fun encode(value: Any): String {
         return (value as InetAddress).hostAddress
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ cache:
 script:
 #  - ./gradlew build -Pnetty_version=4.0.56.Final
 #  - ./gradlew clean build -x test
-#  - ./gradlew :mysq-async:test --tests com.github.jasync.sql.db.mysql.QueryTimeoutSpec --info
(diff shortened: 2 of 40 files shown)


oshai:
Thanks for the effort! I guess you saw conversation in #131 
In any case a suggestion for competability or migration for users will be also beneficial.

bakatz:
@oshai No worries! Yep, I saw the discussion. In the event that I get this working I'll bump the version to 2.0.0 since this would be a breaking change. Will add some notes to the README as well in that case RE: how to migrate from 1.x to 2.x.

db-async-common/src/main/java/com/github/jasync/sql/db/column/InetAddressEncoderDecoder.kt
```diff
@@ -1,18 +1,10 @@
 package com.github.jasync.sql.db.column
 
-import sun.net.util.IPAddressUtil.textToNumericFormatV4
```
bakatz: This was not compiling for me using JDK 11 (AdoptOpenJDK, windows), and generally using sun.* packages is discouraged - decided to replace with standard java library calls. Let me know if there's a reason this textToNumeric() call was made in the first place - seems like .getByName() works OK

oshai: I think I checked it and this method not exists in java 8

bakatz: Yep, the textToNumeric methods aren't public/don't exist or something. That's why I decided to remove it.

bakatz:
hey @oshai I made some progress and fixed all of the common tests. However, I'm running into a strange exception with the MySQL tests. It seems one test is failing due to:

```
com.github.jasync.sql.db.mysql.PreparedStatementsSpec > read a timestamp with microseconds FAILED
    java.util.concurrent.ExecutionException at PreparedStatementsSpec.kt:277
        Caused by: java.lang.UnsupportedOperationException
```

Does this sound familiar to you? I haven't been able to figure out what's causing it.

Thanks

oshai:
Try to run the test seperatly and provide full stacktrace. Right now it doesn't looks familiar.

oshai:
@bakatz - whats the status of this PR? are you still planning on working on it?

bakatz:
@oshai not exactly - haven't had time to work on this due to work commitments - feel free to take the PR over if you have time

oshai:
@bakatz I don't have the time to do that as well, just thinking how we can continue that effort.

PR #159, opened 2019-10-19
R2dbc rc2

### LINKED ISSUE #152: Upgrade to R2DBC 0.8.0.RC1
R2DBC 0.8.0.RC1 is [scheduled for the end of September](https://github.com/r2dbc/r2dbc-spi/milestone/6). It makes sense to take a look at most recent SPI snapshots to catch up with the SPI changes.

oshai:
fixes #152 
Upgraded directly to RC2 - skipped RC1

### DIFF
--- a/r2dbc-mysql/src/main/java/JasyncClientConnection.kt
+++ b/r2dbc-mysql/src/main/java/JasyncClientConnection.kt
@@ -21,15 +21,9 @@ class JasyncClientConnection(
     private val mySQLConnectionFactory: MySQLConnectionFactory
 ) : Connection, Supplier<JasyncConnection> {
 
-    override fun isAutoCommit(): Boolean {
-        TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
-    }
+    private var autoCommit = true
 
     override fun getTransactionIsolationLevel(): IsolationLevel {
-        TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
-    }
-
-    override fun setAutoCommit(autoCommit: Boolean): Publisher<Void> {
         TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
     }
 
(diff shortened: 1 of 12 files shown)


r2dbc-mysql/src/main/java/JasyncClientConnection.kt
```diff
@@ -21,6 +47,15 @@ class JasyncClientConnection(private val jasyncConnection: JasyncConnection) : C
         return executeVoid("COMMIT")
     }
 
+    override fun isAutoCommit(): Boolean {
+        return autoCommit
+    }
+
+    override fun setAutoCommit(autoCommit: Boolean): Publisher<Void> {
+        this.autoCommit = autoCommit
```
mp911de: Setting auto-commit state should be deferred (i.e. on completion or on subscribe of the `SET AUTOCOMMIT` statement)

r2dbc-mysql/src/main/java/JasyncClientConnection.kt
```diff
@@ -41,9 +76,14 @@ class JasyncClientConnection(private val jasyncConnection: JasyncConnection) : C
     }
 
     override fun setTransactionIsolationLevel(isolationLevel: IsolationLevel): Publisher<Void> {
+        this.isolationLevel = isolationLevel
```
mp911de: Setting should be deferred. See auto-commit comment.

r2dbc-mysql/src/main/java/JasyncClientConnection.kt
oshai: I changed it in latest commit, thanks!

r2dbc-mysql/src/main/java/JasyncClientConnection.kt
oshai: I couldn't find this information on the server status so kept it as is.

PR #179, opened 2020-01-15
change prepared statement id from atomic counter to UUID

asockalingam:


asockalingam:
There is no functional change. Existing test cases should pass.

oshai:
Seems like there is a failing test in the build.

asockalingam:
added a fix.

PR #197, opened 2020-05-10
fix security issue

asockalingam:


oshai:
@asockalingam the solution you suggested is not working (causing stackoverflow) see here for alternatives: https://mustafaali.net/2018/01/14/Kotlin-data-classes-and-sensitive-information/

PR #199, opened 2020-05-12
fix security issue

asockalingam:


db-async-common/src/main/java/com/github/jasync/sql/db/ConnectionPoolConfiguration.kt
```diff
@@ -118,6 +118,16 @@ data class ConnectionPoolConfiguration @JvmOverloads constructor(
         coroutineDispatcher = coroutineDispatcher
     )
 
+    override fun toString() = "ConnectionPoolConfiguration(host=$host, port=REDACTED, database=$database, " +
+            "username=REDACTED, password=REDACTED, maxActiveConnections=$maxActiveConnections, " +
+            "maxIdleTime=$maxIdleTime, maxPendingQueries=$maxPendingQueries, " +
+            "connectionValidationInterval=$connectionValidationInterval, " +
+            "connectionCreateTimeout=$connectionCreateTimeout, connectionTestTimeout=$connectionTestTimeout,\n" +
+            "queryTimeout=$queryTimeout, eventLoopGroup=$eventLoopGroup, " +
```
x0a1b, line 126: Why do we need to stringify eventLoopGroup

db-async-common/src/main/java/com/github/jasync/sql/db/ConnectionPoolConfiguration.kt
```diff
@@ -118,6 +118,16 @@ data class ConnectionPoolConfiguration @JvmOverloads constructor(
...
+            "executionContext=$executionContext, coroutineDispatcher=$coroutineDispatcher, ssl=$ssl\n" +
```
x0a1b, line 127: Same is true for `executionContext` and `coroutineDispatcher`

db-async-common/src/main/java/com/github/jasync/sql/db/ConnectionPoolConfiguration.kt
```diff
@@ -118,6 +118,16 @@ data class ConnectionPoolConfiguration @JvmOverloads constructor(
         coroutineDispatcher = coroutineDispatcher
     )
 
+    override fun toString() = "ConnectionPoolConfiguration(host=$host, port=REDACTED, database=$database, " +
```
x0a1b, line 121: You can use multiline string feature here

```kotlin
"""
...
"""
```

db-async-common/src/main/java/com/github/jasync/sql/db/ConnectionPoolConfiguration.kt
asockalingam: fixed.

db-async-common/src/main/java/com/github/jasync/sql/db/ConnectionPoolConfiguration.kt
asockalingam: fixed.

Issue #201, opened 2020-05-24
Convert Gradle to Kotlin script


gumimin:
I did a task converting Gradle to Kotlin script written in [TODO](https://github.com/jasync-sql/jasync-sql/blob/master/the-old-files/TODO.md).
Can I submit a PR?
Changes are pushed in my forked repository.
https://github.com/gumimin/jasync-sql/tree/convert-gradle-to-kts

oshai:
I probably have to update the TODO list... Such PR is welcome!

PR #202, opened 2020-05-24
Convert Gradle to Kotlin script

gumimin:
## Issue
https://github.com/jasync-sql/jasync-sql/issues/201


oshai:
@gumimin thanks for the effort! I will start review it soon. After I finish I will try to release artifacts from this branch to make sure everything is working for release as well.

oshai:
@gumimin two questions:
1. why did you rename to jacocoTestReport? any specific issue with it?
2. about ktlint, it adds a lot of files to the PR, I wonder if it's better to move it to another PR. In addition, I think that it's better to add it to the build in such a way that it will fail the build if linter is failing.

gumimin:
@oshai 
Thank you for your reviewing!
I'm glad to contribute this repository.

---

> 1. why did you rename to jacocoTestReport? any specific issue with it?

Because of failed to create coverage reports, I changed to 'jacocoTestReport' which is the default name of jacoco.
I have no other reasons, I'll restore to 'codeCoverageReport'. ;)

> 2. about ktlint, it adds a lot of files to the PR, I wonder if it's better to move it to another PR. In addition, I think that it's better to add it to the build in such a way that it will fail the build if linter is failing.

You got it. I agree to create another PR and add ktlint check when the build.

---

I also have two questions.

1. To rollback above changes, which is better to use
  - git revert
  - git resset and force push 

2. Which is better to include checking lint at the build time into this PR or the ktlint PR?

oshai:
I think for the rollback do whatever more convinient, anyway I can squash all commits later if needed.
The lint checks should be in the ktlint PR.

gumimin:
@oshai 
I fixed all of above.
To run ktlint checks after merging this PR, I also fixed the both build.gradle.kts in this PR and build.gradle in the other PR.
- https://github.com/jasync-sql/jasync-sql/pull/202/files#diff-392475fdf2bc320d17762ed97109a121R64-R66
- https://github.com/jasync-sql/jasync-sql/pull/204/files#diff-c197962302397baf3a4cc36463dce5eaR70

oshai:
@gumimin after merging ktlint PR, this one has conflicts. If you can resolve them it will be great (I can't push to the PR) or else we can use #205 that I created.
In addition, I noticed you changed some dependencies from compile to implementation. This might cause compatibility issues for users, please set it back as it was in the groovy files. I don't want to mix changes with the move to kotlin gradle. We can rethink those later if required.
Let me know if there are any other changes you did and aware of.

gumimin:
@oshai 
Thank you for merging ktlint PR.

I changed some dependencies from compile to implementation because compile is already deprecated.
I agree to your review and restored this changes.

Currently, changes in this PR are: 
- convert from build.gradle to build.gradle.kts
- convert from settings.gradle to settings.gradle.kts
- update Gradle from v4.9 to v6.4.1 (because Gradle started to support Kotlin DSL from v5.0)
- move versions from versions.gradle to gradle.properties
- replace the ktlint with a plugin
- add the ktlintCheck to the default Gradle check task
- replace the javadocJar task with withJavadocJar()
- replace the sourcesJar task with withSourcesJar()
- move sourceSets extensions from each subprojects' build.gradle to allprojects in build.gradle.kts

oshai:
@gumimin - thanks for the detailed change list!

I tried cloning the branch and releasing a version from it and had some issues:

1. Versions were not configured correctly. I had to move it to `allprojects` block like this:
```
allprojects {
    group = "com.github.jasync-sql"
    version = "1.1.1-kts-8"

    repositories {
        mavenCentral()
        jcenter()
    }

    apply(plugin = "kotlin")
    apply(plugin = "maven-publish")
    apply(plugin = "jacoco")

    java.sourceCompatibility = JavaVersion.VERSION_1_8
    java.targetCompatibility = JavaVersion.VERSION_1_8
...
```
I saw that you moved it and wondered if there was a reason to do so.

2. Got a NPE in publish and it turned out to be this issue: https://github.com/bintray/gradle-bintray-plugin/issues/267

Fixed it by downgrading bintray plugin to 1.8.3.

3. Got an error message and failure to publish to maven central: `Repository name, package name or version name are null for project: root project 'jasync-sql'`

I think it is related to this issue: https://github.com/bintray/gradle-bintray-plugin/issues/279

Maybe because it is in the plugins block and not like it was configured before in the `buildscript` block, but I'm not sure how to configure that in `kts`.

The bottom line is that currently I'm stuck only on the third issue (and checking the artifacts) before merging this PR.

gumimin:
@oshai 
I fixed errors occurred in publish without the last issue. Please check it out.
I'm sorry but it's difficult to find the solution of the last issue because I'm not able to try to publish.
Both of the build.gradle and build.gradle.kts are able to use the plugin block without the buildscript block.
I think current plugin block looks like correctly.

Perhaps, it will be work apply the maven-publish plugin in the subprojects block.
Please try to write ``` apply(plugin = "maven-publish") ``` at L:80 in build.gradle.kts.

oshai:
I tried what you suggest but it didn't help :-(

oshai:
I did some more experiments with no success. However, from inspecting the [code](https://github.com/bintray/gradle-bintray-plugin/blob/5f6a3bd0fb619dc156ad2cfe9cc4add1c8ed4db2/src/main/groovy/com/jfrog/bintray/gradle/tasks/BintrayUploadTask.groovy#L232) that produces this warning it looks pretty safe and that it actually works as expected (the root project should not upload any artifact).

oshai:
I created version `1.1.2`. 
Let's wait for a week or so to give people chance to use it and report if there is any issue, and if all ok we will merge this PR.

gumimin:
Thank you for your efforts and releasing the [v1.1.2](https://bintray.com/jasync-sql/jasync-sql/jasync-sql/1.1.2).

> Let's wait for a week or so to give people chance to use it and report if there is any issue, and if all ok we will merge this PR.

This sounds good. I'll check too.

PR #207, opened 2020-07-01
add quill to 'who is using it' in readme.md

gabfssilva:
Since version 3.5.2, both mysql and postgresql modules were added to Quill: https://github.com/getquill/quill#quill-jasync


oshai:
Thanks!

PR #209, opened 2020-09-05
fix #184 Implement SuspendingConnectionPool

### LINKED ISSUE #184: Add useSuspending to ConnectionPool for better Kotlin support
The current implementation of `.use` method is based on `CompletableFuture` and is not Kotlin-friendly. Instead, I had to write my own implementation that looks more or less like this:
```
suspend fun <T, C : ConcreteConnection> ConnectionPool<C>.useSuspending(
    block: suspend (SuspendingConnection) -> T
): T {
    var exception: Throwable? = null
    val connection = take().await()
    try {
        return block(connection.asSuspending)
    } catch (e: Throwable) {
        exception = e
        throw e
    } finally {
        if (exception == null) {
            giveBack(connection).await()
        } else {
            try {
                giveBack(connection).await()
            } catch (closeException: Throwable) {
                exception.addSuppressed(closeException)
            }
        }
    }
}
```
However, unlike the already provided `.use` function, this implementation is not using `configuration.executionContext`. Not sure how much of a problem this is.
At any rate, it would be nice to have a similar API available for Kotlin users.

P.S. I initially started with r2dbc interface for jasync, but the code ends up bloated with all of the "await" statements...

ilya40umov:
This is a draft version of `SuspendingConnectionPool` which is more in line with `SuspendingConnection` implementation (compared to what I originally suggested in #184). If you like the idea I'll add some tests / write some java docs.

BTW I'm not entirely sure I like the idea of creating another coroutine scope for the sake of being able to delegate to the underlying `ConnectionPool.use` implementation. Since `use` is a suspending function it will already have a coroutine scope it can run everything in, including callback. But then again, it might be safe to create another coroutineScope to make sure that uncaught errors aren't propagated to the CoroutineExceptionHandler of the scope the function will be called in...

### DIFF
--- a/db-async-common/src/main/java/com/github/jasync/sql/db/pool/SuspendingConnectionPool.kt
+++ b/db-async-common/src/main/java/com/github/jasync/sql/db/pool/SuspendingConnectionPool.kt
@@ -1,0 +1,61 @@
+package com.github.jasync.sql.db.pool
+
+import com.github.jasync.sql.db.ConcreteConnection
+import com.github.jasync.sql.db.Connection
+import com.github.jasync.sql.db.QueryResult
+import com.github.jasync.sql.db.SuspendingConnection
+import com.github.jasync.sql.db.asSuspending
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.future.await
+import kotlinx.coroutines.future.future
+
+fun <T : ConcreteConnection> ConnectionPool<T>.asSuspending(): SuspendingConnectionPool<T> =
+    SuspendingConnectionPool(this)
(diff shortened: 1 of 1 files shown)


oshai:
The code looks good, waiting for tests and docs for a full review. 
One idea that can be added:
`Connection.asSuspending` (in SuspendingConnection file) can be enhanced with an `if` to support both regular connection and connection pool. In addition to that we can change SuspendingConnection to have ConcreteConnection instead of Connection as property and delete the `if` in `SuspendingConnectionImpl.inTransaction()` as it will never be a pool then. I think it will make code more readable and direct users to use the correct implementation. You can decide if you want to add it or leave it like you did.

PR #214, opened 2020-11-20
Add client SSL configuration

### LINKED ISSUE #213: Support client SSL (mutual auth)
Right now ssl config allows for a cert from the server and sslmode, but doesn't support the client authenticating via a cert and private key. Some db configurations require the client to authenticate as well

rockwotj:
This is a minimal change to introduce client SSL configuration to the postgres driver.

This additionally upgrades the testcontainers dependency because it wouldn't start up otherwise on MacOS.

Fixes #213


oshai:
Some mysql tests are failing. I suspect it is related to the testcontainers upgrade but need to further check that.

oshai:
Unfortunately I can't run tests locally with the old version, probably due to this issue: https://github.com/testcontainers/testcontainers-java/issues/1623
I will try to check it on travis.

rockwotj:
> Unfortunately I can't run tests locally with the old version, probably due to this issue: [testcontainers/testcontainers-java#1623](https://github.com/testcontainers/testcontainers-java/issues/1623)
> I will try to check it on travis.

That was the issue I had to upgrade for. I can also debug tomorrow morning

oshai:
I tried without test containers upgrade and it seems to be working: https://travis-ci.org/github/jasync-sql/jasync-sql/builds/745282931

rockwotj:
@oshai I got it working. It was an annoying behavior change from this PR: https://github.com/testcontainers/testcontainers-java/pull/857

The fix was to override configure and change it back to "test".

Also the "create the docker image during class initialization" is a not fun pattern to debug... Would you be willing to accept some more PRs to clean that up and instead use J[Unit's builtin ExternalResource](https://junit.org/junit4/javadoc/latest/org/junit/rules/ExternalResource.html)?

oshai:
@rockwotj thanks for all the effort!
Another PR with external resource is welcome. 
Before we merge this, Please note that some mysql tests are still failing in CI.

rockwotj:
> @rockwotj thanks for all the effort!
> Another PR with external resource is welcome.
> Before we merge this, Please note that some mysql tests are still failing in CI.

Oops, I was just running the connection spec, now they should pass now I've fixed the default config values.

oshai:
Thanks @rockwotj !

PR #219, opened 2021-01-10
Add scram-sha-256 authentication (via SASL) to the PostgreSQL client

dustin-johnson:
PostgreSQL added support for [scram-sha-256](https://www.2ndquadrant.com/en/blog/password-authentication-methods-in-postgresql) in [version 10](https://www.postgresql.org/docs/10/sasl-authentication.html). This new mechanism is quite a bit more security conscious than its predecessor MD5, and its use is rapidly increasing. Without support for scram-sha-256, jasync-sql is incompatible with any PostgreSQL server that has it enabled.

Note: I've tried to break this PR up into logical commits to be clear about the intent behind the changes. I think this will be especially helpful in this PR, as there were a number of changes necessary to make the development environment suitable for Windows.


db-async-common/src/main/java/com/github/jasync/sql/db/column/InetAddressEncoderDecoder.kt
```diff
@@ -1,20 +1,8 @@
...
-        } else {
-            InetAddress.getByAddress(textToNumericFormatV4(value))
-        }
-    }
-
-    override fun encode(value: Any): String {
-        return (value as InetAddress).hostAddress
-    }
+    override fun decode(value: String): Any = InetAddress.getByName(value)
+    override fun encode(value: Any): String = (value as InetAddress).hostAddress
```
dustin-johnson, line 7: textToNumericFormatV4/V6 appear to be inaccessible. I'm not sure why this wasn't building on my setup, but I figured it was probably best to migrate off of references to the sun.* package.

db-async-common/src/main/java/com/github/jasync/sql/db/util/BufferDumper.java
```diff
@@ -16,7 +16,7 @@ public static String dumpAsHex(ByteBuf buffer) {
         StringBuilder outputBuf = new StringBuilder(length * 4);
 
         int p = 0;
-        int rows = length / 8;
+        int rows = (length + 4) / 8 - 1;
```
dustin-johnson, line 19: Minor: prevent an empty final row from being printed in some cases.

gradle.properties
```diff
@@ -2,22 +2,20 @@ KOTLIN_VERSION=1.4.20
...
+SCRAM_CLIENT_VERSION=1.0.0-beta.2
 R2DBC_SPI_VERSION=0.8.0.RELEASE
 REACTOR_CORE_VERSION=3.2.6.RELEASE
-
 JUNIT_VERSION=4.12
 ASSERTJ_VERSION=3.11.1
 MOCKK_VERSION=1.9
 LOGBACK_VERSION=1.1.8
-TEST_CONTAINERS_VERSION=1.15.0
+TEST_CONTAINERS_VERSION=1.15.1
```
dustin-johnson, line 19: This new version [fixes](https://github.com/testcontainers/testcontainers-java/issues/3574) a Windows compatibility issue.

gradle.properties
```diff
@@ -2,22 +2,20 @@ KOTLIN_VERSION=1.4.20
...
-
 JACOCO_VERSION=0.8.3
-
 KOTLIN_COROUTINES_VERSION=1.4.2
 SL4J_VERSION=1.7.25
 JODA_VERSION=2.9.7
 JODA_CONVERT_VERSION=1.8.1
 NETTY_VERSION=4.1.49.Final
 KOTLIN_LOGGING_VERSION=1.7.6
+SCRAM_CLIENT_VERSION=1.0.0-beta.2
```
dustin-johnson, line 12: Even though this is a beta version, it's the [version used by the official PostgreSQL JDBC driver](https://github.com/ahachete/pgjdbc/blob/f3ade07d7758f134fd137db3ebab544f2211e290/pgjdbc/pom.xml#L43) and it's the latest version that exists.

gradlew.bat
```diff
@@ -1,84 +1,84 @@
...
-rem Set variable GRADLE_EXIT_CONSOLE if you need the _script_ return code instead of
-rem the _cmd.exe /c_ return code!
-if  not "" == "%GRADLE_EXIT_CONSOLE%" exit 1
-exit /b 1
-
-:mainEnd
-if "%OS%"=="Windows_NT" endlocal
-
-:omega
+@if "%DEBUG%" == "" @echo off
```
dustin-johnson, line 1: Correct file line endings.

mysql-async/src/test/java/com/github/jasync/sql/db/mysql/ConnectionPoolConfigurationSpec.kt
```diff
@@ -8,7 +8,7 @@ import org.junit.Test
 class ConnectionPoolConfigurationSpec : ConnectionHelper() {
 
     @Test
-    fun `"configured connection pool" should "be able to run a query"`() {
+    fun `configured connection pool should be able to run a query`() {
```
dustin-johnson, line 11: Functions with quotes in their name causes compilation issues on Windows.

postgresql-async/build.gradle.kts
```diff
@@ -24,10 +25,12 @@ dependencies {
     compile("io.netty:netty-transport:$NETTY_VERSION")
     compile("io.netty:netty-handler:$NETTY_VERSION")
     compile("io.github.microutils:kotlin-logging:$KOTLIN_LOGGING_VERSION")
+    compile("com.ongres.scram:client:$SCRAM_CLIENT_VERSION")
     testImplementation("junit:junit:$JUNIT_VERSION")
     testImplementation("org.jetbrains.kotlin:kotlin-test-junit:$KOTLIN_VERSION")
     testImplementation("org.jetbrains.kotlin:kotlin-test-junit:$KOTLIN_VERSION")
     testImplementation("org.assertj:assertj-core:$ASSERTJ_VERSION")
+    testImplementation("org.jetbrains.kotlin:kotlin-reflect:$KOTLIN_VERSION")
```
dustin-johnson, line 33: Solve a compiler warning about kotlin-reflect being 1.3.x while the rest of kotlin support is 1.4.x. Note, this warning appears for other projects (such as mysql's client) as well but I didn't want to be presumptuous in proliferating this fix.

postgresql-async/src/main/java/com/github/jasync/sql/db/postgresql/PostgreSQLConnection.kt
```diff
@@ -273,20 +286,42 @@ class PostgreSQLConnection @JvmOverloads constructor(
...
                 this.authenticated = true
             }
-            is AuthenticationChallengeCleartextMessage
-            -> {
-                write(this.credential(message))
+            is AuthenticationCleartextPasswordMessage -> write(this.createPasswordMessage(message))
+            is AuthenticationMD5PasswordMessage -> write(this.createPasswordMessage(message))
+            is AuthenticationSASLMessage -> {
+                // Configuration mirrors PostgreSQL's JDBC driver:
+                // https://github.com/ahachete/pgjdbc/blob/master/pgjdbc/src/main/java/org/postgresql/jre8/sasl/ScramAuthenticator.java
```
dustin-johnson, line 298: As this comment indicates, I leaned heavily on the JDBC implementation for reference. The PR for that change is here, if you'd like to review it as well: https://github.com/pgjdbc/pgjdbc/pull/842/files

postgresql-async/src/main/java/com/github/jasync/sql/db/postgresql/util/PasswordHelper.java
```diff
@@ -26,29 +24,19 @@ private static void bytesToHex(byte[] bytes, byte[] hex, int offset) {
 
             i += 1;
         }
-
     }
 
     public static byte[] encode(String userText, String passwordText, byte[] salt, Charset charset) throws NoSuchAlgorithmException {
```
dustin-johnson, line 29: There are no functional changes in this file, just cleanups/simplifications.

postgresql-async/src/test/java/com/github/aysnc/sql/db/integration/PostgreSQLConnectionSpec.kt
```diff
@@ -79,7 +79,7 @@ class PostgreSQLConnectionSpec : DatabaseTestHelper() {
             time_column time,
             boolean_column boolean,
             constraint bigserial_column_pkey primary key (bigserial_column)
-          ) with oids"""
+          )"""
```
dustin-johnson, line 82: Support for this query syntax appears to have been removed in later version is PostgreSQL: https://www.postgresql.org/message-id/E1gPG3j-0003cy-03%40gemulon.postgresql.org

dustin-johnson:
The above failing builds are obviously concerning. Are they related to my changes? If so, any tips on what might be wrong?

oshai:
@dustin-johnson since the PR is quite big I suggest the following: separate it into two PR's, first one with all the commits that do not include `scram-sha-256 authentication` addition and on top of it the addition itself. splitting to commits is good, but I don't think it allows looking at the tests separately (tests only run on the last commit).

Some more comments from initial review:
- Please upgrade postgres in tests to the minimum version required - so we know we support all those versions. Is it 10 in this case?
- Can you give more details on https://github.com/ongres/scram ? are there any other alternatives? I am a bit hesitant in adding dependencies to the driver.
- The tests run with Java 8. I suggest you make sure they passed before doing any code changes (like InetAddress etc').

dustin-johnson:
> separate it into two PR's, first one with all the commits that do not include scram-sha-256 authentication addition and on top of it the addition itself. splitting to commits is good, but I don't think it allows looking at the tests separately (tests only run on the last commit).

I will look into how to do this without losing the comments on this PR.

> Please upgrade postgres in tests to the minimum version required - so we know we support all those versions. Is it 10 in this case?

I can certainly do that, but I would think we'd want to demonstrate support for the latest released version of Postgres so it doesn't look like this is an aging project. Postgres 10 was released in 2017. I'll wait until I hear back from you.

> Can you give more details on https://github.com/ongres/scram ? are there any other alternatives? I am a bit hesitant in adding dependencies to the driver.

I understand the desire to keep dependencies to a minimum. I can offer these points as consolation: 1 This is the dependency that the official PostgreSQL JDBC driver uses, as described above (with links if you're interested). 2 This dependency doesn't have any dependencies of it's own. 3 It's a security library, so we gain credibility by using the same library as the other implementation do. 4 It's a security library, so we should be hesitant to reimplement this from scratch.

> The tests run with Java 8. I suggest you make sure they passed before doing any code changes (like InetAddress etc').

Did they pass on master? I can look into this a bit more.

oshai:
I merged #220 , please try to rebase that one so we will see less changes.

> I can certainly do that, but I would think we'd want to demonstrate support for the latest released version of Postgres so it doesn't look like this is an aging project. Postgres 10 was released in 2017. I'll wait until I hear back from you.

I think maintaining compatibility is important so let's try postgres 10. Altough you can already drop features that are not supported in newer versions like `oid`.

> I understand the desire to keep dependencies to a minimum. I can offer these points as consolation: 1 This is the dependency that the official PostgreSQL JDBC driver uses, as described above (with links if you're interested). 2 This dependency doesn't have any dependencies of it's own. 3 It's a security library, so we gain credibility by using the same library as the other implementation do. 4 It's a security library, so we should be hesitant to reimplement this from scratch.

SGTM. Let's keep the dependency.

> Did they pass on master? I can look into this a bit more.

Yes, tests pass on master.

dustin-johnson:
- The PostgreSQL integration tests have been downgraded to use version 10.
- This branch has been rebased onto the latest master (and now the build passed even though there was no material changes).

I believe this PR is ready for re-review.

postgresql-async/src/main/java/com/github/jasync/sql/db/postgresql/PostgreSQLConnection.kt
```diff
@@ -273,20 +286,42 @@ class PostgreSQLConnection @JvmOverloads constructor(
...
-            is AuthenticationChallengeMD5
-            -> {
-                write(this.credential(message))
+            is AuthenticationSASLFinalMessage -> {
+                try {
+                    (scramClientFinalProcessor
+                        ?: throw AuthenticationException("Received a SASL final message before the continue message"))
+                        .receiveServerFinalMessage(message.saslData)
+                } catch (e: ScramException) {
+                    throw AuthenticationException("Server failed SCRAM validation")
```
oshai, line 323: don't you want to add e as cause?

postgresql-async/src/main/java/com/github/jasync/sql/db/postgresql/encoders/PasswordEncoder.kt
```diff
@@ -0,0 +1,62 @@
...
+import com.github.jasync.sql.db.postgresql.messages.frontend.SASLInitialResponse
+import com.github.jasync.sql.db.postgresql.messages.frontend.SASLResponse
+import com.github.jasync.sql.db.postgresql.util.PasswordHelper
+import com.github.jasync.sql.db.util.ByteBufferUtils
+import com.github.jasync.sql.db.util.XXX
+import io.netty.buffer.ByteBuf
+import io.netty.buffer.Unpooled
+import java.nio.charset.Charset
+
+class PasswordEncoder(val charset: Charset) : Encoder {
```
oshai, line 15: why did you rename that? it came out as a delete/add so hard to compare (and track history later)

postgresql-async/src/main/java/com/github/jasync/sql/db/postgresql/exceptions/MissingCredentialInformationException.kt
```diff
@@ -1,16 +1,10 @@
...
-    val password: String?,
-    private val authenticationResponseType: AuthenticationResponseType
+    val password: String?
 ) : DatabaseException(
-    "Username and password were required by auth type %s but are not available (username=<%s> password=<%s>".format(
-        authenticationResponseType,
-        username,
-        password
-    )
+    "Username and password were required but are not available (username=<$username> password=<$password>)"
```
oshai, line 9: I know it was like that before, but maybe it's better not to put the password to the message? (maybe only first two chars?)

oshai (commented):
Thanks for all the effort on this!
I left some comments. Think it's mostly minor issues, haven't spotted anything major as much as I could follow.

gradle.properties
dustin-johnson: Wow, that's embarrassing. I was looking at the fork instead of the official JDBC driver. Very good catch!

postgresql-async/src/main/java/com/github/jasync/sql/db/postgresql/PostgreSQLConnection.kt
dustin-johnson: Sounds good. Added.

postgresql-async/src/main/java/com/github/jasync/sql/db/postgresql/encoders/PasswordEncoder.kt
dustin-johnson: The intent was to reduce confusion. It's previous name was CredentialEncoder, but it really only handled passwords. Now that we have another authentication mechanism (SCRAM-SHA-256) that also deals with credentials, I thought it would be clearer to name distinguish this class from the other forms of authentication.

postgresql-async/src/main/java/com/github/jasync/sql/db/postgresql/exceptions/MissingCredentialInformationException.kt
dustin-johnson: Completely agree. I've pushed a fix that leaves the password out entirely while still giving an indication if a password was provided.

dustin-johnson:
I believe I have corrected (or responded to) all feedback. Thank you for the thorough review.

oshai:
Thanks again for the PR!

dustin-johnson:
Fantastic. Thanks for accepting the new feature. What are your thoughts on timing for the next release?

PR #222, opened 2021-01-30
Support SSL for MySQL

rockwotj:
This PR adds support for the ssl configuration in the MySQL driver. There is also a refactoring to keep the SSL related code shared between the drivers (because the drivers are doing the same START_TLS mechanism).


oshai:
Thanks for the PR! It will be great for the review if you can add your reference for the protocol and details on the change as much as possible.

rockwotj:
The SSL guide in the protocol can be found here: https://dev.mysql.com/doc/internals/en/ssl.html

We add a new message for SSLRequest that we send if the config requests SSL and the server supports it. Once we get the handshake from the server we send this request then upgrade to a SSL connection, then send the handshake response as usual.

The SSLRequest message is the same as the handshake response message, just truncated after the padding. I refactored the handshake response message to share that code. I'm happy to change this bit up if you'd rather not share that bit (this was inspired from the way other drivers model this), but the spec doesn't talk about this "header" concept so I can see the case for sharing this serialization in another way.

db-async-common/src/main/java/com/github/jasync/sql/db/util/NettyUtils.kt
```diff
@@ -41,6 +49,38 @@ object NettyUtils {
         else -> NioSocketChannel::class.java
     }
 
+    fun createSslContext(sslConfiguration: SSLConfiguration): SslContext {
```
rockwotj, line 52: All the changes in this file just pull out the logic from the postgres driver because it's the exact same for MySQL SSL

mysql-async/src/main/java/com/github/jasync/sql/db/mysql/MySQLConnection.kt
```diff
@@ -246,17 +251,68 @@ class MySQLConnection @JvmOverloads constructor(
...
+            SSLConfiguration.Mode.Require,
+            SSLConfiguration.Mode.VerifyCA,
+            SSLConfiguration.Mode.VerifyFull -> {
+                require(message.supportsSSL()) { "SSL is not supported on server" }
+                true
+            }
+        }
+
+        val sslRequest = SSLRequestMessage(setOfNotNull(
+                CapabilityFlag.CLIENT_PLUGIN_AUTH,
```
rockwotj, line 266: These flags need to be the same for both messages, instead of taking this message as an argument we can also just share the flag set.

mysql-async/src/main/java/com/github/jasync/sql/db/mysql/MySQLConnection.kt
```diff
@@ -246,17 +251,68 @@ class MySQLConnection @JvmOverloads constructor(
...
+            database = configuration.database,
+            password = configuration.password,
+            appName = configuration.applicationName
         )
+
+        if (!switchToSsl) {
+            connectionHandler.write(handshakeResponse)
+            return
+        }
+
```
rockwotj, line 291: This is the actual changes to the protocol, hopefully it reads like the spec.

mysql-async/src/main/java/com/github/jasync/sql/db/mysql/util/CapabilityFlag.kt
```diff
@@ -0,0 +1,14 @@
+@file:Suppress("unused")
+
+package com.github.jasync.sql.db.mysql.util
+
+enum class CapabilityFlag(val value: Int) {
```
rockwotj, line 5: I think this is a more typesafe change and makes it easier to reason about these ints in the system.

mysql-async/src/main/java/com/github/jasync/sql/db/mysql/encoder/HandshakeResponseEncoder.kt
```diff
@@ -4,60 +4,26 @@ import com.github.jasync.sql.db.exceptions.UnsupportedAuthenticationMethodExcept
...
-        if (m.database != null) {
-            clientCapabilities = clientCapabilities or CLIENT_CONNECT_WITH_DB
-        }
-
-        if (m.appName != null) {
-            clientCapabilities = clientCapabilities or CLIENT_CONNECT_ATTRS
-        }
-
-        val buffer = ByteBufferUtils.packetBuffer()
+        val buffer = headerEncoder.encode(m.header)
```
rockwotj, line 25: Let me know if you'd like a different way to share this serialization logic.

rockwotj (commented):
Some quick comments on the code itself to hopefully help with reviewing

oshai:
Thanks! I will try to review it as soon as I can. Except some delay because of the situation.

rockwotj:
@oshai just a friendly ping to keep this near the top of your notifications :)

oshai:
Yes, sorry, it's on my list. I'll try to do that soon.

mysql-async/src/main/java/com/github/jasync/sql/db/mysql/MySQLConnection.kt
```diff
@@ -246,17 +251,68 @@ class MySQLConnection @JvmOverloads constructor(
...
+        val channelFuture = connectionHandler.write(sslRequest)
+        channelFuture.addListener { sslRequestFuture ->
+            if (!sslRequestFuture.isSuccess) return@addListener
```
oshai, line 294: what will happen in case of failure? maybe add log warning?

mysql-async/src/main/java/com/github/jasync/sql/db/mysql/MySQLConnection.kt
```diff
@@ -246,17 +251,68 @@ class MySQLConnection @JvmOverloads constructor(
...
+            val channel = channelFuture.channel()
+            val handler = try {
+                val sslContext = NettyUtils.createSslContext(configuration.ssl)
+                val sslEngine = sslContext.newEngine(channel.alloc(), configuration.host, configuration.port)
+                if (configuration.ssl.mode == SSLConfiguration.Mode.VerifyFull) {
+                    NettyUtils.verifyHostIdentity(sslEngine)
+                }
+                SslHandler(sslEngine)
+            } catch (e: Exception) {
+                setException(e)
```
oshai, line 304: it's better to also log the exception (even with debug log) for cases that the future is not logging it from some reason.

mysql-async/src/main/java/com/github/jasync/sql/db/mysql/MySQLConnection.kt
```diff
@@ -246,17 +251,68 @@ class MySQLConnection @JvmOverloads constructor(
...
+                return@addListener
+            }
+            channel.pipeline().addFirst(handler)
+            handler.handshakeFuture().addListener { handshakeFuture ->
+                if (handshakeFuture.isSuccess) {
+                    connectionHandler.write(handshakeResponse)
+                } else {
+                    handshakeFuture.cause()?.let(::setException)
```
oshai, line 312: same here about error logging

mysql-async/src/main/java/com/github/jasync/sql/db/mysql/util/CapabilityFlag.kt
oshai: great!

mysql-async/src/test/java/com/github/jasync/sql/db/mysql/ContainerHelper.java
```diff
@@ -1,13 +1,13 @@
 package com.github.jasync.sql.db.mysql;
 
-import com.github.jasync.sql.db.Configuration;
-import com.github.jasync.sql.db.Connection;
-import com.github.jasync.sql.db.QueryResult;
-import com.github.jasync.sql.db.RowData;
+import com.github.jasync.sql.db.*;
```
oshai, line 3: please don't add star imports

oshai (commented):
Thanks for the hard work! overall looks good, left some minor comments

mysql-async/src/main/java/com/github/jasync/sql/db/mysql/MySQLConnection.kt
rockwotj: I've added a comment, but `connectionHandler.write` is in charge of errors in this case.

mysql-async/src/main/java/com/github/jasync/sql/db/mysql/MySQLConnection.kt
rockwotj: Done.

mysql-async/src/main/java/com/github/jasync/sql/db/mysql/MySQLConnection.kt
rockwotj: Done.

mysql-async/src/test/java/com/github/jasync/sql/db/mysql/ContainerHelper.java
rockwotj: Sorry IntelliJ does this by default. Reverted.

oshai:
Thanks for the PR!

PR #228, opened 2021-02-21
Fix #226 - don't log password

### LINKED ISSUE #226: PostgreSQLConnectionFactory logs eintire db configuration including password
The PostgreSQLConnectionFactory the entire config on DEBUG level:

```
    init {
        logger.debug { "PostgreSQLConnectionFactory created with configuration $configuration" }
    }
```

Either the toString function should possibly be overwritten to not log sensitive data or the log message should be removed. A workaround would be to create your own connection factory class.

oshai:


db-async-common/src/main/java/com/github/jasync/sql/db/Configuration.kt
```diff
@@ -55,3 +55,10 @@ data class Configuration @JvmOverloads constructor(
     val eventLoopGroup: EventLoopGroup = NettyUtils.DefaultEventLoopGroup,
     val executionContext: Executor = ExecutorServiceUtils.CommonPool
 )
+
+fun Configuration.toDebugString(): String {
+    val toString = this.toString()
+    return toString.replace(passRegex, ", password=****, database=")
+}
+
+private val passRegex = ", password=.*, database=".toRegex()
```
rockwotj, line 64: You probably want this to be non-greedy, otherwise it's possible that an awful applicationName breaks this.

oshai: Thanks! I changed the implementation to copy data instead of non-greedy regexp.

rockwotj (approved):

Issue #229, opened 2021-03-09
BinaryRowEncoder does not support encode java.time.LocalDateTime


ONI-MIKO:
**environment**

jasync-r2dbc-mysql version: 1.1.7

```xml
<!-- r2dbc mysql -->
<dependency>
   <groupId>com.github.jasync-sql</groupId>
   <artifactId>jasync-r2dbc-mysql</artifactId>
   <version>1.1.7</version>
</dependency>
```

**insert into database report this error:**

```
Caused by: java.lang.UnsupportedOperationException: An operation is not implemented: couldn't find mapping for class java.time.LocalDateTime
	at com.github.jasync.sql.db.mysql.binary.BinaryRowEncoder.encoderFor(BinaryRowEncoder.kt:100) ~[jasync-mysql-1.1.7.jar:na]
	at com.github.jasync.sql.db.mysql.encoder.PreparedStatementExecuteEncoder.encodeValue(PreparedStatementExecuteEncoder.kt:62) ~[jasync-mysql-1.1.7.jar:na]
	at com.github.jasync.sql.db.mysql.encoder.PreparedStatementExecuteEncoder.encodeValues(PreparedStatementExecuteEncoder.kt:46) ~[jasync-mysql-1.1.7.jar:na]
	at com.github.jasync.sql.db.mysql.encoder.PreparedStatementExecuteEncoder.encode(PreparedStatementExecuteEncoder.kt:27) ~[jasync-mysql-1.1.7.jar:na]
	at com.github.jasync.sql.db.mysql.codec.MySQLOneToOneEncoder.encode(MySQLOneToOneEncoder.kt:67) ~[jasync-mysql-1.1.7.jar:na]
	at com.github.jasync.sql.db.mysql.codec.MySQLOneToOneEncoder.encode(MySQLOneToOneEncoder.kt:23) ~[jasync-mysql-1.1.7.jar:na]
	at io.netty.handler.codec.MessageToMessageEncoder.write(MessageToMessageEncoder.java:89) ~[netty-codec-4.1.58.Final.jar:4.1.58.Final]
	... 33 common frames omitted
```

**this is my entity example**

```java
import java.time.LocalDateTime;

public class MyEntity {
    /**
     * 创建时间
     */
    private LocalDateTime createTime;
    /**
     * 更新时间
     */
    private LocalDateTime updateTime;
}
```

**I saw this part of the code for the *BinaryRowEncoder***

```kotlin
[...]

PR #233, opened 2021-05-28
Switch to JSR 310 Date and Time API

dragneelfps:
- Removes joda from encoding/decoding

- [x] common
- [x] mysql
- [x] postgres

continuation of https://github.com/jasync-sql/jasync-sql/issues/229


dragneelfps:
Getting following in CI build.. :/ 
```
docker: Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit.
```

oshai:
Hi, thanks for the PR!
There was some interest for it in the past. For reference see #131 and #157.
Ideally, if we could have both implementation as a non-breaking change it would be preferred.
But I understand this might complicate things a lot.
Since this is a breaking change we will need to bump the major version when merged.

dragneelfps:
I could try to have Joda Time supported as well. But the thing is, do we really want to continue supporting it?
I prefer the major version upgrade. But seems  oddlike going from `1.1.7` to `2.0.0`.
Suggestions on how to move forward with this please?

oshai:
Bump version to 2.0.0 sounds reasonable to me.
Please rebase changes from master again, build should work now in CI.

oshai:
Note that some tests are still failing on CI.

dragneelfps:
I fixed most. Just having issues understanding logic behind 
https://github.com/jasync-sql/jasync-sql/blob/152cdef93d98ca137b424e6dda389ce7304d02f7/postgresql-async/src/test/java/com/github/aysnc/sql/db/integration/TimeAndDateSpec.kt#L151
can you help?

oshai:
> I fixed most. Just having issues understanding logic behind
> https://github.com/jasync-sql/jasync-sql/blob/152cdef93d98ca137b424e6dda389ce7304d02f7/postgresql-async/src/test/java/com/github/aysnc/sql/db/integration/TimeAndDateSpec.kt#L151
> 
> 
> can you help, @oshai ?

I didn't wrote it originally, but from reading the code I think it checks the following:
- set timestamp that includes mills/micros
- check that the timestamp millis conversion is greater than the second before and smaller from the second after

There might be a problem either with the timezone or the millis if I have to guess.

dragneelfps:
I noticed something, postgres is returning timestamps in UTC string, even when I insert it with non UTC offset. 
The tests are expecting non UTC offset. So I changed a bit and is actually testing if the two timestamps are equal as OffsetDateTime
Is this correct approach?
I am guessing postgres returning UTC string is related to server settings (which I am not really changing). Is this correct assumption?

oshai:
> I noticed something, postgres is returning timestamps in UTC string, even when I insert it with non UTC offset.
> The tests are expecting non UTC offset. So I changed a bit and is actually testing if the two timestamps are equal as OffsetDateTime
> Is this correct approach?
> I am guessing postgres returning UTC string is related to server settings (which I am not really changing). Is this correct assumption?

According to https://www.postgresql.org/docs/9.1/datatype-datetime.html 8.5.2. Date/Time Output:
```
The default is the ISO format
and it looks like:

ISO | ISO 8601/SQL standard | 1997-12-17 07:37:16-08

(without time zone)
```

Conversion on test makes sense (to see it's the same date as inserted). It's just not clear how this was changed from pervious impl.

I still need to review all changes, but it will take some time.

mysql-async/src/main/java/com/github/jasync/sql/db/mysql/binary/encoder/DateTimeEncoder.kt
```diff
@@ -2,14 +2,13 @@ package com.github.jasync.sql.db.mysql.binary.encoder
...
 import com.github.jasync.sql.db.mysql.column.ColumnTypes
 import io.netty.buffer.ByteBuf
-import org.joda.time.LocalDateTime
-import org.joda.time.ReadableDateTime
+import java.time.OffsetDateTime
 
 object DateTimeEncoder : BinaryEncoder {
     override fun encode(value: Any, buffer: ByteBuf) {
-        val instant = value as ReadableDateTime
+        val zdt = value as OffsetDateTime
```
oshai, line 9: isn't it more suitable to use ZonedDateTime here?

mysql-async/src/test/java/com/github/jasync/sql/db/mysql/ConnectionHelper.kt
```diff
@@ -66,8 +66,8 @@ open class ConnectionHelper : ContainerHelper() {
         """CREATE TEMPORARY TABLE posts (
        id INT NOT NULL AUTO_INCREMENT,
        created_at_date DATE not null,
-       created_at_datetime DATETIME not null,
-       created_at_timestamp TIMESTAMP not null,
+       created_at_datetime DATETIME(6) not null,
```
oshai, line 69: why was it changed?

oshai (commented):
Thanks again for the PR.
I left some comments in code and have some more general things:
- I saw that in some cases there is no test coverage: do you think it's an issue with the framework, or can you please add tests to cover those cases?
- I saw in many places move from millis to nano. Is it because java 8 uses nano and joda/db use millis?

About the migration to 2.x: I am thinking if we can add a suggestion to users how to do that more easily? Do you have any suggestion?

mysql-async/src/main/java/com/github/jasync/sql/db/mysql/binary/encoder/DateTimeEncoder.kt
dragneelfps: Apparently, it is recommended to be used when storing to database. 
doc: https://docs.oracle.com/javase/8/docs/api/java/time/OffsetDateTime.html
A good explanation: https://stackoverflow.com/a/30234992/5375462

gradle.properties
```diff
@@ -19,3 +19,4 @@ LOGBACK_VERSION=1.1.8
 TEST_CONTAINERS_VERSION=1.15.1
 MYSQL_CONNECTOR_VERSION=5.1.47
 AWAITILITY_VERSION=3.1.5
+THREETEN_EXTRA=1.6.0
```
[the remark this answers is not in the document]
dragneelfps: In java.time, we have two APIs - Duration(seconds and nanoseconds) and Period(years, months and days).
For joda, the Period API supports (seconds and nanoseconds) & (years, months and days).
For users, who are already using joda Period do denote a full period, we need a replacement. And here comes the https://www.threeten.org/threeten-extra/apidocs/org.threeten.extra/org/threeten/extra/PeriodDuration.html

db-async-common/src/main/java/com/github/jasync/sql/db/column/LocalDateTimeEncoderDecoder.kt
```diff
@@ -1,27 +1,25 @@
...
 
     private const val ZeroedTimestamp = "0000-00-00 00:00:00"
 
-    private val optional = DateTimeFormatterBuilder()
-        .appendPattern(".SSSSSS").toParser()
-
     private val format = DateTimeFormatterBuilder()
         .appendPattern("yyyy-MM-dd HH:mm:ss")
-        .appendOptional(optional)
+        .append(microsecondsFormatter)
```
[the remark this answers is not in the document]
dragneelfps: When using joda date time formatter, a `SSSSSS` pattern will match all `9`, `99`, `999`, etc.
But java cannot. It is strict in nature. for pattern  `SSSSSS`, tt requires it to have 6 `S`s in the date time string, otherwise it will throw an error.
Thats why I am using a `microsecondsFormatter` which supports all possible count of `S`, which is 1-6.

dragneelfps:
> I saw that in some cases there is no test coverage: do you think it's an issue with the framework, or can you please add tests to cover those cases?

Can you point me to which cases are you talking about specifically?

> I saw in many places move from millis to nano. Is it because java 8 uses nano and joda/db use millis?

Yes. Java uses nano, joda uses millis, db micro(at least in mysql)

oshai:
> Can you point me to which cases are you talking about specifically?

Look at LocalDateTimeEncoderDecoder:17 and see more examples in 'files changed' with the warning: `Added line #Lxx was not covered by tests`

mysql-async/src/main/java/com/github/jasync/sql/db/mysql/binary/encoder/DateTimeEncoder.kt
oshai: I was mostly referring [this](https://blog.joda.org/2014/11/converting-from-joda-time-to-javatime.html) that said this is the equivalent for joda DateTime.

gradle.properties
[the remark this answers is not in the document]
oshai: ok

db-async-common/src/main/java/com/github/jasync/sql/db/column/LocalDateTimeEncoderDecoder.kt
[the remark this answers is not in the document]
oshai: ok, but isn't the fact that it's optional means it can be absent (no digit or dot at all)?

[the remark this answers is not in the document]
dragneelfps: right. I think I might have made it `append` instead of `appendOptional` by mistake. will correct it

mysql-async/src/main/java/com/github/jasync/sql/db/mysql/binary/encoder/DateTimeEncoder.kt
dragneelfps: the doc you referenced states both are acceptable, no?

oshai: yes, only that `ZonedDateTime` is semantically the same as `DateTime` while the other is similar.

dragneelfps: well I could change it to Zoned but even official java doc says to use Offset for saving into database and thats what https://github.com/pgjdbc/r2dbc-postgresql/ does for reference.

oshai: ok, let's keep it with OffsetDateTime

db-async-common/src/main/java/com/github/jasync/sql/db/column/LocalDateTimeEncoderDecoder.kt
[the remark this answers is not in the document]
dragneelfps: fixed

oshai:
@dragneelfps before merging there was still one answered question in the review about `DATETIME(6)`.

In addition, will you have some time to improve test coverage?

oshai:
I will merge this PR soon. Will create 1.x branch before that.

oshai:
Thanks for the PR!
I merged it and will release a new version and update docs soon.

PR #242, opened 2021-08-06
Allow connections in a pool to be initialized with a user-provided SQL query

dustin-johnson:
Add connectionInitializationQuery as a new parameter to ConnectionPoolConfiguration. The new parameter enables users to provide an SQL query that will be invoked on all connections created for the pool. This can be useful for ensuring state tied to the connection lifecycle is properly configured, such as search_path for PostgreSQL.

The motivation for this change is closely related to this previous request from an unrelated user: https://gitter.im/jasync-sql/support?at=6039e0d0e562cf54ac87aa57

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ project/target/*
 project/project/*
 postgresql-async/target/*
 db-async-common/target/*
+pool-async/target/*
 db-async-common/out/*
 mysql-async/out/*
 postgresql-async/out/*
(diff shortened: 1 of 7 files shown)


dustin-johnson:
I believe the changes are fully tested, but the tests are written as integration tests. Perhaps the integration tests don't count toward code coverage?

oshai:
Hi, thanks for the PR!
Coverage should work for integration tests as well, so I am not sure why it's dropped.
About the change itself. I think maybe a better approach is to allow users to provide a different connection factory?

It will be more flexible like that.
It's also pretty easy to "work-around" that by implementing your own `createConnectionPool()`

I also don't think it should be additional configuration for the pool, maybe just another parameter to `createConnectionPool()`?

dustin-johnson:
I agree that it would be a lot more flexible to provide your own connection factory. The big downside is that the connection factories are extensions of the `ConnectionFactory` abstract class. Do we want to make `ConnectionFactory` a part of the API by encouraging users to extend it and then manually instantiating `ConnectionPool`? If that's the case than I can decline this PR and simply use this approach for my use case.

dustin-johnson:
On this point, perhaps we could make use of the `internal` keyword to be explicit.

PR #247, opened 2021-08-23
add mysql CLIENT_FOUND_ROWS support

oshai:
This commit adds system property jasync.mysql.CLIENT_FOUND_ROWS that
when set will create connections with the flag on.

CLIENT_FOUND_ROWS tells the mysql server to send back all rows in where
clause and not just the ones that were actually updated. See details in:
https://dev.mysql.com/doc/refman/8.0/en/information-functions.html
https://dev.mysql.com/doc/internals/en/capability-flags.html

This makes more sense in some scenarios. For example:
https://github.com/spring-projects/spring-data-r2dbc/issues/253

Also renamed SSLRequestMessage to CapabilityRequestMessage to better
reflect the data.

### DIFF
--- a/mysql-async/src/main/java/com/github/jasync/sql/db/mysql/MySQLConnection.kt
+++ b/mysql-async/src/main/java/com/github/jasync/sql/db/mysql/MySQLConnection.kt
@@ -61,6 +61,7 @@ class MySQLConnection @JvmOverloads constructor(
         @Suppress("unused")
         val MicrosecondsVersion = Version(5, 6, 0)
         private val regexForCallInQueryStart = Regex("\\s*call\\s+.*", RegexOption.IGNORE_CASE)
+        const val CLIENT_FOUND_ROWS_PROP_NAME = "jasync.mysql.CLIENT_FOUND_ROWS"
     }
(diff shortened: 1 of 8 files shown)


mysql-async/src/main/java/com/github/jasync/sql/db/mysql/MySQLConnection.kt
```diff
@@ -262,8 +262,14 @@ class MySQLConnection @JvmOverloads constructor(
             }
         }
 
-        val sslRequest = SSLRequestMessage(setOfNotNull(
+        val clientFoundRows = System.getProperty("jasync.mysql.CLIENT_FOUND_ROWS") != null
```
ks-yim, line 265: Thanks for the prompt fix.
So, capability can be enabled with JVM option after this PR.

I think it would be nicer if it detects `mysql-r2dbc` dependency and automatically enable the flag?

oshai: good idea, I will do something like that on the connection factory.

PR #260, opened 2021-09-06
fix bug connection fault tolerance

wind-hx:
Multiple connection error handling, Test passed.


mysql-async/src/main/java/com/github/jasync/sql/db/mysql/MySQLConnection.kt
```diff
@@ -113,6 +113,9 @@ class MySQLConnection @JvmOverloads constructor(
     override fun lastException(): Throwable? = this.lastException
 
     override fun connect(): CompletableFuture<MySQLConnection> {
+        if (this.connectionHandler.isConnected()) {
+            return this.connectionPromise
```
oshai, line 117: can you add a test to cover that? e.g. calling connect twice?

oshai (commented):
Thanks for the PR!

PR #283, opened 2022-01-16
Yomanz master

oshai:


Yomanz:
Will it be tagged after this PR?

oshai:
added in 2.0.6.

Yomanz:
> added in 2.0.6.

I don't see it on maven.

PR #290, opened 2022-04-28
Fix connections leaking on high load.

sbmpost:
1. Moves the responsibility of connection create timeout completely to the connection.
2. By doing so, allows us to close the underlying Netty Channel if the creation times out.

### DIFF
--- a/postgresql-async/src/test/java/com/github/aysnc/sql/db/integration/PostgreSQLPoolConfigurationSpec.kt
+++ b/postgresql-async/src/test/java/com/github/aysnc/sql/db/integration/PostgreSQLPoolConfigurationSpec.kt
@@ -89,10 +89,7 @@ class PostgreSQLPoolConfigurationSpec : DatabaseTestHelper() {
             "jdbc:postgresql://$host:$port/$database?user=$username&password=$password"
         }
 
-        val connection = PostgreSQLConnectionBuilder.createConnectionPool(connectionUri) {
-            connectionCreateTimeout = 1
-        }
-        assertThat(connection.configuration.connectionCreateTimeout).isEqualTo(1)
+        val connection = PostgreSQLConnectionBuilder.createConnectionPool(connectionUri)
         try {
             return fn(connection)
         } finally {


postgresql-async/src/main/java/com/github/jasync/sql/db/postgresql/codec/PostgreSQLConnectionHandler.kt
```diff
@@ -94,22 +94,49 @@ class PostgreSQLConnectionHandler(
...
                                 when (ty2) {
                                     is Success -> this.disconnectionPromise.success(this)
                                     is Failure -> this.disconnectionPromise.failed(ty2.exception)
                                 }
                             }
                         }
-                        is Failure -> this.disconnectionPromise.failed(ty1.exception)
+                        is Failure -> {
+                            println("<<<<<<< should close channel (write close message failure)")
+//                            closeChannel()
```
sbmpost, line 133: TODO: How to close the channel if write fails?

postgresql-async/src/main/java/com/github/jasync/sql/db/postgresql/codec/PostgreSQLConnectionHandler.kt
```diff
@@ -94,22 +94,49 @@ class PostgreSQLConnectionHandler(
         return this.connectionFuture
     }
 
+/*
```
sbmpost, line 97: Not sure yet if we are going to need this

sbmpost:
@oshai 
I have completed my PR. It would be great if you can have a look.

db-async-common/src/main/java/com/github/jasync/sql/db/Configuration.kt
```diff
@@ -50,6 +51,7 @@ data class Configuration @JvmOverloads constructor(
     val maximumMessageSize: Int = 16777216,
     val allocator: ByteBufAllocator = PooledByteBufAllocator.DEFAULT,
     val connectionTimeout: Int = 5000,
+    val createTimeout: Duration = Duration.ofMillis(5000),
```
oshai, line 54: I think this is exactly as connectionTimeout

mysql-async/src/test/java/com/github/jasync/sql/db/MySQLSlowConnectionDelegate.kt
```diff
@@ -0,0 +1,43 @@
...
+    private val onOkSlowdownInMillis: Long
+) : MySQLHandlerDelegate {
+    override fun onHandshake(message: HandshakeMessage) =
+        delegate.onHandshake(message)
+
+    override fun onError(message: ErrorMessage) =
+        delegate.onError(message)
+
+    override fun onOk(message: OkMessage) {
+        Thread.sleep(onOkSlowdownInMillis)
```
oshai, line 22: what is this for?

mysql-async/src/test/java/com/github/jasync/sql/db/MySQLSlowConnectionDelegate.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.github.jasync.sql.db.mysql.message.server.HandshakeMessage
+import com.github.jasync.sql.db.mysql.message.server.EOFMessage
+import com.github.jasync.sql.db.mysql.message.server.ErrorMessage
+import com.github.jasync.sql.db.mysql.message.server.OkMessage
+import io.netty.channel.ChannelHandlerContext
+
+class MySQLSlowConnectionDelegate(
+    private val delegate: MySQLHandlerDelegate,
+    private val onOkSlowdownInMillis: Long
+) : MySQLHandlerDelegate {
```
oshai, line 14: kotlin has a delegate feature, maybe you can use that

pool-async/src/main/java/com/github/jasync/sql/db/pool/ActorBasedObjectPool.kt
```diff
@@ -337,17 +336,6 @@ private class ObjectPoolActor<T : PooledObject>(
         }
     }
 
-    private fun checkItemsInCreationForTimeout() {
```
oshai, line 340: I would keep that as a safety mechanism, maybe with a bigger threshold like createTimeout * 2 (or + 20 seconds)

oshai (commented):
Thanks for the PR!

db-async-common/src/main/java/com/github/jasync/sql/db/Configuration.kt
sbmpost: Indeed, will fix this

mysql-async/src/test/java/com/github/jasync/sql/db/MySQLSlowConnectionDelegate.kt
sbmpost: That certainly removed a lot of boilerplate code, thanks!

pool-async/src/main/java/com/github/jasync/sql/db/pool/ActorBasedObjectPool.kt
sbmpost: Sounds like a good idea to keep things backwards compatible. I have restored it.

sbmpost:
@oshai 
I have gone over all review comments and made the necessary changes.

sbmpost:
@oshai can we consider to merge this?

oshai:
The CI fails on formatting (also failed on using old kotlin version without last ,). I will fix it later today or you can do that - I want to make sure CI pass before merging.

oshai:
The change cause some tests fail. Maybe the timeout is misconfigured or there is another issue. I would appreciate if you can take a look.

sbmpost:
@oshai 
Previously the connectionTimeout setting was passed to the NIO java socket library which due to its non-blocking nature never really times out. But now that we set up our own timer as well in order to properly cleanup the channel before the pool forceably destroys the connection (with createTimeout=2*connectionTimeout), suddenly the connectionTimeout actually times out. This however made a test fail that demanded a 1 second connectionTimeout. By removing this 1 second check, the test now succeeds.

sbmpost:
@oshai I had to fix the same thing for postgresql. I can't restart the test myself, but I think all should be well now.

oshai:
ok thanks! I will merge once all tests pass.

oshai:
Thanks for the PR! I will create a new version soon.

PR #304, opened 2022-09-18
update r2dbc to 1.0.0

oshai:


r2dbc-mysql/src/main/java/JasyncInsertSyntheticMetadata.kt
```diff
@@ -29,12 +28,12 @@ internal class JasyncInsertSyntheticMetadata(private val generatedKeyName: Strin
         return this
     }
 
-    override fun getColumnMetadatas(): Iterable<ColumnMetadata> {
-        return listOf(this)
+    override fun getColumnMetadatas(): MutableList<out ColumnMetadata> {
+        return mutableListOf(this)
     }
 
-    override fun getColumnNames(): Collection<String> {
```
duyleekun, line 36: The removal of `getColumnNames` method is making this backward incompatible with `r2dbc-spi` `0.9.1`. Therefore, `2.0.8` will be the last version that's compatible with `0.9.1`
For user friendly purposes, I guess we should put some compatible matrix in the `README.md`

oshai: I added a section in the wiki, let me know if you think it's enough: https://github.com/jasync-sql/jasync-sql/wiki/FAQ#q-is-r2dbc-version--1-supported

duyleekun: What about back port fix for 0.9.1 or it's not supported anymore?

oshai: the interface is in spi and it was changed, so there is no way to backport it.

duyleekun: Since the latest stable spring boot `2.7.4` is still on spi `0.9.2`. In my opinion, it may take a lot of time for people to move to spring boot `3.0.0`, especially since it's still in `3.0.0-M5` right now. Just stating my opinion, I wouldn't know your opinion on this though.

oshai: @mp911de do you know what is the timeline for spring released with r2dbc 1.0?

oshai: I saw that RC was published and GA is planned for next month, so I think our best option is to wait for that release.

duyleekun: Yeah. I'm coming here from spring 2.x documentation, so it caught me by surprise that it doesn't work out of the box. I guess we can notify the Spring team or modify the library README.md file instead of hiding it inside the wiki.

oshai: I added a section to the readme, hope it's clear enough: https://github.com/jasync-sql/jasync-sql/blob/master/README.md#compatibility

mp911de: Spring Framework 6 GA is scheduled for mid-November. As you've noticed, RC1 releases are already available.

Issue #312, opened 2022-09-27
r2dbc interface unable to close prepared statement


knuclechan:
I've found a previous [thread](https://github.com/jasync-sql/jasync-sql/issues/244) of the same issue, but it is unhandled and closed. However, it is a very serious issue that should be handled. I've spend a lot of time of investigating why the mysql server crash periodically, and it is not easy to find out it is due to unclosed prepared statements. I don't think I am the only one. It is very likely to happen if you are using connection pool and have queries that have arbitrary number of items in an `in` condition.

I think the library should provide a mean for the user to specify if the prepared statement should be close or not. If the library are not going to provide such a choice. Please just set the default to close the prepared statement, instead of setting up a trap that crashes production mysql server after up for some weeks and potentially corrupting the data.

I would suggest:
1. Add a configuration to connectionFactory to config the default behavior.
2. Allow setting the behavior per query in the r2dbc.spi.Statement object.
3. Better set the default behavior to close the prepared statement

oshai:
I see that [r2dbc](https://github.com/r2dbc/r2dbc-pool) now has `maxLifeTime`, so one option is using it.   

The driver has support to close prepared statement, see #82, but I think r2dbc don't have such api.  

So one approach would be to open an issue in `r2dbc.spi`.

What we can do in the driver is having the behavior controlled via env prop (ie: `-DautocloseStatements=true`).

oshai:
Added in 2.1.9, thanks!

PR #330, opened 2022-11-09
Add option to r2dbc to close the prepared statement

steveannett:
This pull request would add an option to `JasyncStatement` in jasync-r2dbc-mysql to release the prepared statement after use. By default the prepared statement is left open, but sometimes this is not always the desired option, as reported in https://github.com/jasync-sql/jasync-sql/issues/312


steveannett:
It should be noted I also submitted a feature request in the `r2dbc-spi` library https://github.com/r2dbc/r2dbc-spi/issues/275

r2dbc-mysql/src/main/java/JasyncStatement.kt
```diff
@@ -26,6 +26,7 @@ internal class JasyncStatement(private val clientSupplier: Supplier<JasyncConnec
 
     private var isPrepared = false
     private var selectLastInsertId: Boolean = false
+    private var releaseAfterwards: Boolean = false
```
oshai, line 29: maybe call it releasePreparedStatementAfterUse (and also the method)?

r2dbc-mysql/src/main/java/JasyncStatement.kt
```diff
@@ -71,6 +72,11 @@ internal class JasyncStatement(private val clientSupplier: Supplier<JasyncConnec
         return this
     }
 
+    fun releaseAfterwards(): Statement {
+        releaseAfterwards = true
```
oshai, line 76: also assert that this is prepared statement?

r2dbc-mysql/src/main/java/JasyncStatement.kt
```diff
@@ -71,6 +72,11 @@ internal class JasyncStatement(private val clientSupplier: Supplier<JasyncConnec
         return this
     }
 
+    fun releaseAfterwards(): Statement {
```
oshai, line 75: how can this be used? it's not part of the interface. by casting to JasyncStatement?

r2dbc-mysql/src/main/java/JasyncStatement.kt
steveannett: Thanks for the feedback! I've renamed both the variable and method

r2dbc-mysql/src/main/java/JasyncStatement.kt
steveannett: I've added a check that it is a prepared statement, let me know if that is OK, or if you'd prefer it in a different style

r2dbc-mysql/src/main/java/JasyncStatement.kt
steveannett: Yes, that would be the idea, if the programmer was really keen on making sure that the prepared statement is closed they could cast it and request for it to be closed, for example in Java 14:

```
	private void releaseAfterUse(Statement st) {
		if (st instanceof JasyncStatement jst) jst.releasePreparedStatementAfterUse();
	}
```

oshai:
basically LGTM. One more idea about configuring it: how about adding an option to `MysqlConnectionFactoryProvider`?

I am not a spring expert but maybe that is the intended usage for spring? I guess we can have both to allow connection level control as well. WDYT?

r2dbc-mysql/src/main/java/JasyncStatement.kt
oshai: lgtm

oshai:
@steveannett is it ready to be merged?

steveannett:
@oshai yes, ready to be merged, thanks!

Regarding adding an option to `MySqlConnectionFactoryProvider`, its a good idea, I will look into this and see if I can get it added - although in my particular use case I would like to have the connection level control so I can leave it open in certain cases.

Issue #255, opened 2021-09-03
There is a query still being run here with spring @transactional


maksymilian-gasztych:
I'm getting `There is a query still being run here - race -> false` when use spring transactional annotation, and can't seem to find any solution to that on the internet. It happens for any transactional method, but just to give some example I'm sharing the code below - removing transactional solves the problem, but that's not a viable solution.
```
    @Transactional
    fun editPrefectureRegions(editPrefectureRegionsCommand: EditPrefectureRegionsCommand): Mono<Prefecture> {
        if (editPrefectureRegionsCommand.regionIds.isEmpty()) return Mono.error(ApplicationException(PrefectureMustBeBelongToAtLeastOneRegion))
        return Flux.fromIterable(editPrefectureRegionsCommand.regionIds)
            .flatMap {
                regionRepository.existsById(it)
            }
            .collectList()
            .map {
                if (it.contains(false)) throw ApplicationException(RegionIsNotFound)
                else it
            }
            .flatMap {
                prefectureRepository.findById(editPrefectureRegionsCommand.id)
            }
            .switchIfEmpty(Mono.error(ApplicationException(PrefectureIsNotFound)))
            .flatMap {
                prefectureRepository.save(it.copy(regionIds = editPrefectureRegionsCommand.regionIds))
            }
    }
```

#### Versions

- Driver: jasync-r2dbc-mysql:2.0.2
- Database: mysql:5.7 (docker)
- Java: 11
- OS: mac os 

#### Current Behavior

<details>
<summary>Stack trace</summary>

```
com.github.jasync.sql.db.exceptions.ConnectionStillRunningQueryException: <<mysql-connection-7>> - There is a query still being run here - race -> false
	at com.github.jasync.sql.db.mysql.MySQLConnection.validateIsReadyForQuery(MySQLConnection.kt:425)
	Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException: 
Error has been observed at the following site(s):
	|_ checkpoint ⇢ Handler jp.menu.jinrai.controllers.region.PrefectureController#editPrefectureRegions(int, EditPrefectureRegionsRequest) [DispatcherHandler]
Stack trace:
		at com.github.jasync.sql.db.mysql.MySQLConnection.validateIsReadyForQuery(MySQLConnection.kt:425)
		at com.github.jasync.sql.db.mysql.MySQLConnection.sendQueryDirect(MySQLConnection.kt:334)
[...]

PR #343, opened 2022-12-12
r2dbc-mysql queueing execution of queries

FutureGadget:
This is my suggested fix for issue https://github.com/jasync-sql/jasync-sql/issues/255#issuecomment-1336219534.

Since in reactor environment, multiple threads can access to the same connection unintendedly, I added a queue per `JasyncClientConnection`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,11 +41,19 @@ For docs and info see the [wiki](https://github.com/jasync-sql/jasync-sql/wiki).
   <artifactId>jasync-mysql</artifactId>
   <version>2.1.7</version>
 </dependency>
+
 <!-- postgresql -->
 <dependency>
-  <groupId>com.github.jasync-sql</groupId>
-  <artifactId>jasync-postgresql</artifactId>
-  <version>2.1.7</version>
+    <groupId>com.github.jasync-sql</groupId>
+    <artifactId>jasync-postgresql</artifactId>
+    <version>2.1.7</version>
+</dependency>
+
+<!-- r2dbc-mysql -->
+<dependency>
+    <groupId>com.github.jasync-sql</groupId>
+    <artifactId>jasync-r2dbc-mysql</artifactId>
+    <version>2.1.7</version>
 </dependency>
 ```
 
--- a/r2dbc-mysql/build.gradle.kts
+++ b/r2dbc-mysql/build.gradle.kts
@@ -35,6 +35,7 @@ dependencies {
     compile("io.netty:netty-transport:$NETTY_VERSION")
     compile("io.netty:netty-handler:$NETTY_VERSION")
     compile("io.github.microutils:kotlin-logging:$KOTLIN_LOGGING_VERSION")
+    implementation("org.springframework.data:spring-data-r2dbc:1.5.6")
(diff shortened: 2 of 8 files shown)


oshai:
Thanks for the PR!

Can you give some details on the suggested fix?
The test is a really good start.
Having a queue for queries (with scheduler) seems like not the optimal approach.
I am not sure how the driver expected to handle simultaneous queries on the same connection? multiplexing?
Jasync doesn't support multiplexing at the moment, but it seems transactions are not directly related to that.

FutureGadget:
> Thanks for the PR!
> 
> Can you give some details on the suggested fix? The test is a really good start. Having a queue for queries (with scheduler) seems like not the optimal approach. I am not sure how the driver expected to handle simultaneous queries on the same connection? multiplexing? Jasync doesn't support multiplexing at the moment, but it seems transactions are not directly related to that.

@oshai, thank you for taking a look on this PR.
What I tried to fix with this PR was an "unintended" conflict of query on transaction rollback.

What I mean by "unintended" is when you are in a transaction, cancelling the subscription of ongoing query publisher can cause an asynchronous ROLLBACK query to be executed using the same connection.

This behavior is from the following code. The `transactional` method uses `Mono.usingWhen(...,asyncCancel)`, and the `asyncCancel` callback function parameter will make ROLLBACK happen on subscription cancellation.
```java
Mono.usingWhen(Mono.just(it), 
    ignore -> mono,
    this.transactionManager::commit, 
    (res, err) -> Mono.empty(), 
    this.transactionManager::rollback
)
```
(https://github.com/spring-projects/spring-framework/blob/main/spring-tx/src/main/java/org/springframework/transaction/reactive/TransactionalOperatorImpl.java#L81)

It is "unintended" because a user cannot control the ROLLBACK query to be executed after the ongoing query is finished. As the jasync driver validates the connection before sending any query to the database, the above behavior will make jasync driver throw an exception since the ROLLBACK query cannot be executed simultaneously while the other query being executed using the same connection.

In this PR, I added `QueueingExecutionJasyncConnectionAdapter`, which adapts `MySQLConnection` to use a `ArrayBlockingQueue`. The number of simultaneous queries should be at most 2 (the original first query and the following rollback query), so the size of the queue, which is 256, must be enough.

I didn't intend to introduce a behavior which is similar to query multiplexing with this PR.
Therefore, simultaneous queries using a single connection should be best avoided as it was before this PR.

FutureGadget:
> I have few ideas/options to make it simple:
> 
> * on rollback, just close the connection (which I think will implicitly rollback the current transaction)
> * use [queryPromiseReference](https://github.com/jasync-sql/jasync-sql/blob/254c30e91502a02b682f024f4387738abd8aa4a6/mysql-async/src/main/java/com/github/jasync/sql/db/mysql/MySQLConnection.kt#L90) and chain the cancellation to it (if exists)
> 
> What do you think?

Thank you for better suggestions.
I've just tried the second one (chaining query promise) and it worked like a charm.
Please review : https://github.com/jasync-sql/jasync-sql/pull/345
Thank you.

PR #345, opened 2022-12-13
query promise chaining

FutureGadget:
As @oshai suggested at https://github.com/jasync-sql/jasync-sql/pull/343#issuecomment-1347020196, I implemented query chaining. This can solve the problem mentioned in https://github.com/jasync-sql/jasync-sql/pull/343#issuecomment-1346877775 without introducing scheduler and query execution queue.

### DIFF
--- a/mysql-async/src/main/java/com/github/jasync/sql/db/mysql/MySQLConnection.kt
+++ b/mysql-async/src/main/java/com/github/jasync/sql/db/mysql/MySQLConnection.kt
@@ -46,7 +46,6 @@ import io.netty.handler.ssl.SslHandler
 import mu.KotlinLogging
 import java.time.Duration
 import java.util.Optional
-import java.util.concurrent.CancellationException
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.atomic.AtomicLong
 import java.util.concurrent.atomic.AtomicReference
@@ -347,18 +346,6 @@ class MySQLConnection @JvmOverloads constructor(
         return chainQueryPromiseIfExists(query, this::send)
     }
 
-    private fun CompletableFuture<QueryResult>.cleanUpOnCancellation(): CompletableFuture<QueryResult> {
-        return this.handle { result, ex ->
-            // Clear query promise when ongoing query has been cancelled so that this connection can be reused
-            if (ex is CancellationException) {
-                clearQueryPromise()
-                QueryResult(0L, null)
-            } else {
-                result
-            }
-        }
-    }
-
     private fun <T> chainQueryPromiseIfExists(
         param: T,
         func: (T) -> CompletableFuture<QueryResult>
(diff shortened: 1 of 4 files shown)


mysql-async/src/main/java/com/github/jasync/sql/db/mysql/MySQLConnection.kt
```diff
@@ -340,6 +343,25 @@ class MySQLConnection @JvmOverloads constructor(
...
+        return chainQueryPromiseIfExists(query, this::send)
+    }
+
+    private fun <T> chainQueryPromiseIfExists(
+        param: T,
+        func: (T) -> CompletableFuture<QueryResult>
+    ): CompletableFuture<QueryResult> {
+        return try {
+            if (isQuerying()) {
+                queryPromise().get().flatMap { func(param) }
```
oshai, line 355: can you please add a info log here, that we are doing a chaining?

oshai (commented):
generally lgtm.

Initially I thought to do that only for cancellation, but the proposed cl is doing it for all queries.
This looks like a more gracefull way to go, and the commit is cleaner. On the other hand in some edge cases this might hide issues because now it allows trying to send queries in parallel (it will just chain them).

Since it doesn't break any test I think we can go with that. the info log message might assist in case someone see something unexpected and I hope this is enough.

FutureGadget:
> generally lgtm.
> 
> Initially I thought to do that only for cancellation, but the proposed cl is doing it for all queries. This looks like a more gracefull way to go, and the commit is cleaner. On the other hand in some edge cases this might hide issues because now it allows trying to send queries in parallel (it will just chain them).
> 
> Since it doesn't break any test I think we can go with that. the info log message might assist in case someone see something unexpected and I hope this is enough.

Like you said, I also think that this fix is prone to side effects.
I've tested this with r2dbc-pool and it showed some unexpected behaviors.
For example, the first query (not the chained queries but the very first one) got timed out even after the subscription has been cancelled and `ROLLBACK` and `SET AUTOCOMMIT = 1` queries were not executed.

Since what I expected was that once the first query is cancelled, the `ROLLBACK` and `SET AUTOCOMMIT = 1` 
queries should be chained and executed one by one, so that the connection can be reused later.

For that reason, I've implemented and tested your another suggestion, which is close connection on query conflict.
I would be appreciated if you take a look on https://github.com/jasync-sql/jasync-sql/pull/347.
Thank you!

mysql-async/src/main/java/com/github/jasync/sql/db/mysql/MySQLConnection.kt
FutureGadget: Thank you for your review and inspiring comments. I've tested this fix and it has some side effects as you said.
I think https://github.com/jasync-sql/jasync-sql/pull/347 is a better option.

PR #346, opened 2022-12-14
Update netty dependency to 4.1.86Final

kampka:
The current version of netty is rather old and affected by a series of potential CVE.
We should update to the latest version to get those fixed.


oshai:
Thanks!

PR #347, opened 2022-12-14
r2bc-mysql close connection on query conflict

FutureGadget:
This is an implementation of what @oshai suggested in the following comment:
https://github.com/jasync-sql/jasync-sql/pull/343#issuecomment-1347020196

As @oshai mentioned in the above thread, we may take advantage of an implicit rollback when ROLLBACK query conflict with other running queries.

I prefer this version to the second (https://github.com/jasync-sql/jasync-sql/pull/345) one.
This solution is more simple, stable, and less side effects.

Thank you @oshai for a good idea to start with.

### DIFF
--- a/r2dbc-mysql/build.gradle.kts
+++ b/r2dbc-mysql/build.gradle.kts
@@ -35,6 +35,7 @@ dependencies {
     compile("io.netty:netty-transport:$NETTY_VERSION")
     compile("io.netty:netty-handler:$NETTY_VERSION")
     compile("io.github.microutils:kotlin-logging:$KOTLIN_LOGGING_VERSION")
+    implementation("org.springframework.data:spring-data-r2dbc:1.5.6")
     testImplementation("junit:junit:$JUNIT_VERSION")
     testImplementation("org.jetbrains.kotlin:kotlin-test-junit:$KOTLIN_VERSION")
     testImplementation("org.assertj:assertj-core:$ASSERTJ_VERSION")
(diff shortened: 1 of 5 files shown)


oshai:
I had something like this in mind: https://github.com/jasync-sql/jasync-sql/compare/saferollback

Please let me know what you think.

FutureGadget:
> I had something like this in mind: https://github.com/jasync-sql/jasync-sql/compare/saferollback
> 
> Please let me know what you think.

LGTM.
Can you do the same for `setAutoCommit(autoCommit: Boolean)` in `JasyncClientConnection`?
`SET AUTOCOMMIT = 1` will also be sent simultaneously with the same connection after the `ROLLBACK` by the following code when in transaction: https://github.com/spring-projects/spring-framework/blob/main/spring-r2dbc/src/main/java/org/springframework/r2dbc/connection/R2dbcTransactionManager.java#L360
```

oshai:
I don't want to take the credit for all the effort you did on that issue.

So I suggest that you'll take that snippet and create the full pull request if that's ok with you.

FutureGadget:
@oshai I've created a PR in the Spring repository: https://github.com/spring-projects/spring-framework/pull/29703.
That PR is to release connection even after the transaction clean up (rollback, set autocommit = 1,..) fails.

It's not a direct solution because releasing the connection will eventually be failed because of the failed query promise and the number of allocated connections will be decreased, but the r2dbc-pool will reallocate connections after all.

In the mean time, I am working on the jasync-sql code to make the connection valid even after the query cancellation or query timeout occurs by chaining query promise based on what you've showed to me.

To do that, we might add some optional connection property to ignore cancellation (since for now cancelling the promise will cancel the whole chain which makes query chaining useless).

Let me know if you have some idea. 
Btw, I am planning to make a PR in a few days.
Thanks.

oshai:
@FutureGadget Any update?

FutureGadget:
> @FutureGadget Any update?

Let me start working on this again. I had an idea which is adding a connection option to ignore query cancellation. My memory is fading out since it was quite a while ago, so give me some time so that I can finalize my thoughts. 
Thanks.

oshai:
Implemeented in #378 378

PR #349, opened 2023-01-02
SSLMode support for MysqlConnectionFactoryProvider

### LINKED ISSUE #337: SSL Mode support for MySQL
Hello!

**I would like to ask if there's a plan to support `sslMode` configuration for MySQL database?** According to FAQ, currently, only PostgreSQL is supported. I also noticed that `com.github.jasync.sql.db.SSLConfiguration` defaults to `sslMode=disabled` mode for MySQL which leads to problems when the MySQL Server is configured to enforce SSL connections.

I had a look at the implementation details and found that this library integrates with R2DBC SPI (https://r2dbc.io/spec/0.8.1.RELEASE/spec/html/#overview.connection.discovery). The driver will be discovered by R2DBC SPI which will use `com.github.jasync.r2dbc.mysql.MysqlConnectionFactoryProvider#create` to create `io.r2dbc.spi.ConnectionFactory` object:
```kotlin
// from: com.github.jasync.r2dbc.mysql.MysqlConnectionFactoryProvider#create
@Suppress("NULLABILITY_MISMATCH_BASED_ON_JAVA_ANNOTATIONS")
override fun create(connectionFactoryOptions: ConnectionFactoryOptions): JasyncConnectionFactory {
    val configuration = Configuration(
        host = connectionFactoryOptions.getValue(HOST),
        port = connectionFactoryOptions.getValue(PORT),
        username = connectionFactoryOptions.getValue(USER),
        password = connectionFactoryOptions.getValue(PASSWORD)?.toString(),
        database = connectionFactoryOptions.getValue(DATABASE),
        applicationName = connectionFactoryOptions.getValue(APPLICATION_NAME),
        connectionTimeout = connectionFactoryOptions.getValue(CONNECT_TIMEOUT)?.toMillis()?.toInt() ?: 5000,
        queryTimeout = connectionFactoryOptions.getValue(QUERY_TIMEOUT)
    )
    return JasyncConnectionFactory(MySQLConnectionFactory(configuration))
}
```
The above method does not configure `com.github.jasync.sql.db.SSLConfiguration` and thus, the default `sslMode` (`Mode.Disable`) will be used. Even though, the `connectionFactoryOptions` could have the option `ssl` set to `true` or any `sslMode` parameter specified.

I suspect that changing this factory to interpret the `ssl` and `sslMode` parameter of `connectionFactoryOptions` would quickly resolve the issue. It seems that `com.github.jasync.sql.db.mysql.MySQLConnection#onHandshake` already supports SSL.

**I would be happy to contribute and resolve this issue**. But first I would like to make sure:
 - Is my understanding of current implementation described above correct?
 - The ssl option in `connectionFactoryOptions` would return true in case the secure protocol (`r2dbcs`) is used. However how should it handle the combination of secure protocol and `sslMode=disabled`, for example through the following connection string:  `r2dbcs:mysql://example.com?sslMode=disabled`?
 - Should my contribution also support other MySQL parameters. If so, would anyone be able to specify which? I know there are legacy parameters like  'useSSL', 'requireSSL', and 'verifyServerCertificate' which are still accepted but translated into a value for 'sslMode' by MySQL Server. The `com.github.jasync.sql.db.SSLConfiguration` also accepts `sslrootcert`, `sslcert`, `sslkey`.

Thanks!

tkurylek:
This PR fixes #337

### DIFF
--- a/r2dbc-mysql/src/main/java/MysqlConnectionFactoryProvider.kt
+++ b/r2dbc-mysql/src/main/java/MysqlConnectionFactoryProvider.kt
@@ -69,7 +69,8 @@ class MysqlConnectionFactoryProvider : ConnectionFactoryProvider {
             database = connectionFactoryOptions.getValue(DATABASE) as String?,
             applicationName = connectionFactoryOptions.getValue(APPLICATION_NAME) as String?,
             connectionTimeout = (connectionFactoryOptions.getValue(CONNECT_TIMEOUT) as Duration?)?.toMillis()?.toInt() ?: 5000,
-            queryTimeout = connectionFactoryOptions.getValue(STATEMENT_TIMEOUT) as Duration?
+            queryTimeout = connectionFactoryOptions.getValue(STATEMENT_TIMEOUT) as Duration?,
+            ssl = MysqlSSLConfigurationFactory.create(connectionFactoryOptions)
         )
         return JasyncConnectionFactory(MySQLConnectionFactory(configuration))
     }
--- a/r2dbc-mysql/src/main/java/MysqlSSLConfigurationFactory.kt
+++ b/r2dbc-mysql/src/main/java/MysqlSSLConfigurationFactory.kt
@@ -1,0 +1,33 @@
+package com.github.jasync.r2dbc.mysql
+
+import com.github.jasync.sql.db.SSLConfiguration
+import com.github.jasync.sql.db.SSLConfiguration.Mode.Disable
+import com.github.jasync.sql.db.SSLConfiguration.Mode.Prefer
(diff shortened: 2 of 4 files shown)


tkurylek:
@oshai The `MysqlSSLConfigurationFactoryTest` passes locally on my machine (via `./gradlew clean build`). Do you know why the PR fails here: https://github.com/jasync-sql/jasync-sql/actions/runs/3822628348/jobs/6507771842 ?

oshai:
I will check locally once I get to it.

oshai:
I just run it again and it passed. It looks something is flaky, hope will get to investigate it later.

PR #357, opened 2023-01-14
Fix MySQL startup in Testcontainers

KarboniteKream:
## Description
This PR fixes the `MySQLContainer` configuration to have the container successfully start. Testcontainers uses the credentials provided using `withUsername()` and `withPassword()` to connect to the container and confirm it has started. With the current implementation of `configure()`, the state is never correctly set and Testcontainers cannot connect.

## Detailed changes
- Move some environment variables out of `configure()` so Testcontainers is correctly configured
- Bump `testcontainers` to latest version, to make it work on M1 Macs (fixes JNA dependency and adds support for `amd64` images). Closes #324.
- Update `.gitignore` to ignore more test logs

## Notes
The unix socket exposed in #319 does not work correctly under Docker for Mac with VirtioFS enabled, and also prevents container startup due to permission issues. Switching to gRPC FUSE or osxfs is a temporary workaround until it's fixed on Docker side.


oshai:
Thanks for the PR!

PR #358, opened 2023-01-14
Support `caching_sha2_password` authentication mode

### LINKED ISSUE #297: Unknown authentication method
Hey there!
I've come across the same issue that is described in #117 and in a later one. That is when I send a request to MySQL from my service I receive this crappy message "Unknown authentication method -> 'caching_sha2_password'". I've applied the solution offered in that issue and I've looked up in the docs what it means. I've checked the mysqld variables values in the command line and made sure that that variable's value was updated from that options file. Everything looks fine except the fact that it doesn't work! The service still shows that crap, which means your driver doesn't take this variable from mysqld as it's supposed to, but from somewhere else. What's going on?

KarboniteKream:
## Description
This PR implements partial support for `caching_sha2_password` mode. Closes #297.
Specifically, it supports the fast authentication path, while the full authentication path is over SSL only.
See `priority #1` described on [this page](https://dev.mysql.com/doc/dev/mysql-server/latest/page_caching_sha2_authentication_exchanges.html). This should hopefully cover the majority of cases using this authentication mode. I'll try to implement the rest in a separate PR.

## Detailed changes
* Added `CachingSha2PasswordAuthentication` and `Sha256PasswordAuthentication`.
* Added `AuthMoreDataMessage` and the necessary decoders. This is required to switch from fast to full authentication mode when the password is not cached yet on the server.
* Fixed decoder for `AuthenticationSwitchRequest` to consume all bytes.
* Fixed `OldPasswordAuthentication` to handle native authentication handshakes.
* Updated container tests to run against MySQL 8.0

### DIFF
--- a/mysql-async/src/main/java/com/github/jasync/sql/db/mysql/codec/MySQLFrameDecoder.kt
+++ b/mysql-async/src/main/java/com/github/jasync/sql/db/mysql/codec/MySQLFrameDecoder.kt
@@ -3,6 +3,7 @@ package com.github.jasync.sql.db.mysql.codec
 import com.github.jasync.sql.db.exceptions.BufferNotFullyConsumedException
 import com.github.jasync.sql.db.exceptions.NegativeMessageSizeException
 import com.github.jasync.sql.db.exceptions.ParserNotAvailableException
+import com.github.jasync.sql.db.mysql.decoder.AuthMoreDataDecoder
 import com.github.jasync.sql.db.mysql.decoder.AuthenticationSwitchRequestDecoder
 import com.github.jasync.sql.db.mysql.decoder.ColumnDefinitionDecoder
 import com.github.jasync.sql.db.mysql.decoder.ColumnProcessingFinishedDecoder
@@ -39,6 +40,7 @@ class MySQLFrameDecoder(val charset: Charset, private val connectionId: String) 
     private val handshakeDecoder = HandshakeV10Decoder()
     private val errorDecoder = ErrorDecoder(charset)
     private val okDecoder = OkDecoder(charset)
+    private val authMoreDataDecoder = AuthMoreDataDecoder()
     private val columnDecoder = ColumnDefinitionDecoder(charset, DecoderRegistry(charset))
     private val authenticationSwitchRequestDecoder = AuthenticationSwitchRequestDecoder(charset)
     private val rowDecoder = ResultSetRowDecoder()
@@ -159,6 +161,13 @@ class MySQLFrameDecoder(val charset: Charset, private val connectionId: String) 
                             this.okDecoder
                         }
                     }
+                }
+            }
+            ServerMessage.AuthMoreData -> {
+                if (!isInQuery) {
+                    this.authMoreDataDecoder
+                } else {
+                    null
                 }
             }
             else -> {
--- a/mysql-async/src/main/java/com/github/jasync/sql/db/mysql/codec/MySQLFrameDecoder.kt
+++ b/mysql-async/src/main/java/com/github/jasync/sql/db/mysql/codec/MySQLFrameDecoder.kt
@@ -91,7 +91,7 @@ class MySQLFrameDecoder(val charset: Charset, private val connectionId: String) 
             logger.trace {
                 "[connectionId:$connectionId] - Reading message type $messageType - " +
                     "(count=$messagesCount,hasDoneHandshake=$hasDoneHandshake,size=$size,isInQuery=$isInQuery,processingColumns=$processingColumns,processingParams=$processingParams,processedColumns=$processedColumns,processedParams=$processedParams)" +
-                    "\n${BufferDumper.dumpAsHex(slice)}}"
+                    "\n${BufferDumper.dumpAsHex(slice)}"
             }
 
             slice.markReaderIndex()
--- a/mysql-async/src/main/java/com/github/jasync/sql/db/mysql/decoder/AuthenticationSwitchRequestDecoder.kt
+++ b/mysql-async/src/main/java/com/github/jasync/sql/db/mysql/decoder/AuthenticationSwitchRequestDecoder.kt
@@ -3,14 +3,21 @@ package com.github.jasync.sql.db.mysql.decoder
 import com.github.jasync.sql.db.mysql.message.server.AuthenticationSwitchRequest
 import com.github.jasync.sql.db.mysql.message.server.ServerMessage
 import com.github.jasync.sql.db.util.readCString
-import com.github.jasync.sql.db.util.readUntilEOF
 import io.netty.buffer.ByteBuf
+import io.netty.buffer.ByteBufUtil
 import java.nio.charset.Charset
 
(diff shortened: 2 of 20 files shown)


mysql-async/src/test/java/com/github/jasync/sql/db/mysql/StoredProceduresSpec.kt
```diff
@@ -96,7 +96,7 @@ class StoredProceduresSpec : ConnectionHelper() {
             ).get()
             val rows = connection.sendQuery(
                 """
-                  SELECT routine_name FROM INFORMATION_SCHEMA.ROUTINES WHERE routine_name="remTest"
+                  SELECT routine_name as routine_name FROM INFORMATION_SCHEMA.ROUTINES WHERE routine_name="remTest"
```
KarboniteKream, line 99: Workaround for a breaking change in MySQL 8.0. [Reference](https://bugs.mysql.com/bug.php?id=84456)

KarboniteKream:
@oshai Ready for review. Please take a look when you have time.

mysql-async/src/main/java/com/github/jasync/sql/db/mysql/encoder/auth/Sha256PasswordAuthentication.kt
```diff
@@ -0,0 +1,20 @@
...
+
+object Sha256PasswordAuthentication : AuthenticationMethod {
+
+    private val EmptyArray = ByteArray(0)
+
+    override fun generateAuthentication(charset: Charset, password: String?, seed: ByteArray?): ByteArray {
+        return if (password != null) {
+            val bytes = password.toByteArray(charset)
+            val result = ByteArray(bytes.length + 1)
+            System.arraycopy(bytes, 0, result, 0, bytes.length)
```
oshai, line 14: can use bytes.copyInto(result) instead

mysql-async/src/test/resources/update-config.sh
```diff
@@ -2,7 +2,8 @@
 cp -f /docker-entrypoint-initdb.d/ca.pem /var/lib/mysql/ca.pem
 cp -f /docker-entrypoint-initdb.d/server-cert.pem /var/lib/mysql/server-cert.pem
 cp -f /docker-entrypoint-initdb.d/server-key.pem /var/lib/mysql/server-key.pem
-cat /etc/mysql/my.cnf
+cat /etc/my.cnf || true
+cat /etc/mysql/my.cnf || true
 ls -lah /var/lib/mysql
-ls -lah /etc/mysql/conf.d/
-ls -lah /etc/mysql/mysql.conf.d/
+ls -lah /etc/mysql/conf.d/ || true
```
oshai, line 8: why was that added?

oshai (commented):
Thanks for the PR! it's awesome!
Left some few minor comments.
One question about testing - IIUC we now moved to `caching_sha2_password` for all tests (because of the upgrade to mysql8).
Do you think it worth adding tests for all auth methods supported?

mysql-async/src/main/java/com/github/jasync/sql/db/mysql/decoder/AuthenticationSwitchRequestDecoder.kt
```diff
@@ -3,20 +3,14 @@ package com.github.jasync.sql.db.mysql.decoder
...
         val method = buffer.readCString(charset)
-        val bytes: Int = buffer.readableBytes()
-        val terminal = 0.toByte()
-        val salt = if (bytes > 0 && buffer.getByte(buffer.writerIndex() - 1) == terminal) ByteBufUtil.getBytes(
-            buffer,
-            buffer.readerIndex(),
-            bytes - 1
-        ) else ByteBufUtil.getBytes(buffer)
-        return AuthenticationSwitchRequest(method, salt)
+        val seed = buffer.readUntilEOF(charset).toByteArray(charset)
```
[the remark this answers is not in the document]
oshai: this was based on code here: https://github.com/mirromutth/r2dbc-mysql/blob/main/src/main/java/dev/miku/r2dbc/mysql/message/server/ChangeAuthMessage.java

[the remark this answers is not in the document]
KarboniteKream: I see, thanks! Then let me revert this a little bit, to keep the original behavior.

mysql-async/src/main/java/com/github/jasync/sql/db/mysql/codec/MySQLFrameDecoder.kt
```diff
@@ -161,6 +163,13 @@ class MySQLFrameDecoder(val charset: Charset, private val connectionId: String)
                     }
                 }
             }
+            ServerMessage.AuthMoreData -> {
+                if (!isInQuery) {
+                    this.authMoreDataDecoder
+                } else {
+                    null
+                }
```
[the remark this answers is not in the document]
KarboniteKream: Maybe that's not necessary? We already have the trace log on L89, which logs all messages. The messages with type 1 that happen in command phase are various responses to queries, so an extra log here might not have much value.

mysql-async/src/test/resources/update-config.sh
KarboniteKream: The location of configuration files is different between images for 5.7 and 8.0, so I suppress failures with this.

KarboniteKream:
> Do you think it worth adding tests for all auth methods supported?

I'll try to add something to cover all of them. 👍

mysql-async/src/main/java/com/github/jasync/sql/db/mysql/encoder/auth/OldPasswordAuthentication.kt
```diff
@@ -7,12 +7,22 @@ import kotlin.math.floor
...
-            password != null && password.isNotEmpty() -> {
-                newCrypt(charset, password, String(seed, charset))
+            !password.isNullOrEmpty() -> {
+                // The native authentication handshake will provide a 20-byte challenge.
+                // Use the first 8 bytes as the old password authentication challenge.
+                val challenge = if (seed.length == 20) {
+                    seed.copyOf(8)
+                } else {
+                    seed
+                }
```
KarboniteKream, line 23: [Reference](https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_connection_phase_authentication_methods.html#page_protocol_connection_phase_authentication_methods_old_password_authentication)

> If the server announces Native Authentication in the `Protocol::Handshake` packet the client may use the first 8 bytes of its 20-byte auth_plugin_data as input.

oshai:
Let me know once it's ready, thanks!

KarboniteKream:
Ready for another review! I've added a TODO, which I'll resolve in the next PR.

oshai:
LGTM!

oshai:
Thanks for the effort! Released 2.1.12 with the PR.

Issue #359, opened 2023-01-16
2.1.9 Regression, NoSuchMethodError com.github.jasync.sql.db.Configuration


KevinAtSesam:
```
{
  "level": "ERROR",
  "title": "Uncaught exception",
  "origin": "Main",
  "exception": [
    "java.lang.NoSuchMethodError: 
      'com.github.jasync.sql.db.Configuration com.github.jasync.sql.db.Configuration.copy(java.lang.String, java.lang.String, int, java.lang.String, java.lang.String, com.github.jasync.sql.db.SSLConfiguration, java.nio.charset.Charset, int, io.netty.buffer.ByteBufAllocator, int, java.time.Duration, java.lang.String, java.util.List, io.netty.channel.EventLoopGroup, java.util.concurrent.Executor, java.lang.String, java.lang.String)'
  ]
}
```

oshai:
Can you please provide some more details?

KevinAtSesam:
Of course, let me know what you need.

Anyway, we use a series of libraries in conjunction with jasync-sql . For one, we use [scalikejdbc-async](https://github.com/scalikejdbc/scalikejdbc-async) **0.16.0** for writing our queries and such. When we initialize the database, we we a connection pool.
```
AsyncConnectionPool.add(
  postgresConfig.name,
  postgresConfig.url,
  postgresConfig.user,
  postgresConfig.pass,
  asyncConnectionPoolSettings
)

[...]

  lazy val asyncConnectionPoolSettings: AsyncConnectionPoolSettings = AsyncConnectionPoolSettings(
    // Must be lower then the RDS connection limit https://serverfault.com/a/982173
    maxPoolSize = 900,
    // Total amount of requests that can be held before being executed
    maxQueueSize = 10000,
    // Number of milliseconds for which the objects are going to be kept as idle
    maxIdleMillis = 1000L,
    connectionSettings = AsyncConnectionSettings(
      // Time allowed between our platform and setting up a connection to Postgres
      connectTimeout = Some(5.seconds),
      // Time for testing connection pools
      testTimeout = Some(5.seconds)
      // Uncertain if we should use this
      // queryTimeout = Some(20.seconds)
    )
  )
```

My guess is that one of the settings are not properly passed on. 

## Version kerfuffle
We are currently on jasync-postgresql 2.1.8 and scalikejdbc-async 0.16.0, and all works well, but the moment I upgrade the jasync library to 2.1.9, there are errors when starting the database connection.

[...]

PR #362, opened 2023-01-17
fix for #359 Configuration copy binary compatibility

oshai:


oshai:
replaced by https://github.com/jasync-sql/jasync-sql/pull/368

PR #378, opened 2023-01-29
Make transaction cleanup possible on request cancellation

FutureGadget:
Two things happen on request cancellation when using netty based webserver under spring reactive transaction.
1. Rollback query is executed asynchronously using the same connection that the current transaction holds.
2. Clean up queries are executed to revert back to the original state of the connection.

If and only if when the above steps are executed without exception, the connection is properly returned to the pool.
I've made a PR https://github.com/spring-projects/spring-framework/pull/29703 to make sure under any circumstances the connection be returned to the pool.

However, apart from the fix, I think we can improve driver part too.

If this PR is merged, both steps will be successful and the connection will be attempted to be returned to the pool even though it will always fail. After the failure of the attempt, acquired connections will be decreased (I added an integration test for this) and the following requests will be handled properly by other connections in the pool. 

When the number of connections in the pool gets to zero, connections will be reallocated and there is no problem.

I got a lot of help from @oshai specifically by the following PR: https://github.com/jasync-sql/jasync-sql/compare/saferollback.
Thank you.


oshai:
Thanks for the PR!

PR #379, opened 2023-01-30
Fix for TX isolation level issue

turnik:
Fix for the `com.github.jasync.sql.db.mysql.exceptions.MySQLException: Error 1568 - #25001 - Transaction characteristics can't be changed while a transaction is in progress` issue https://github.com/jasync-sql/jasync-sql/issues/380


oshai:
Thanks for the contribution!

PR #384, opened 2023-02-24
test(r2dbc-mysql): add transaction double queries test

ForteScarlet:
test for #383


oshai:
Right now it fails: KtLint found code style violations. You can fix it:
./gradlew ktlintForma

PR #391, opened 2023-04-13
[r2dbc-mysql] Fix `getColumnMetadatas()` of `JasyncMetadata`

### LINKED ISSUE #390: [r2dbc-mysql] `getColumnMetadatas()` of `JasyncMetadata` lose columns with the same names
I use spring-r2dbc and (latest 2.1.23) jasync-r2dbc-mysql for my project, and there are a few inner join queries.

But when I use `databaseClient.sql("~").fetch().all()~` for joining tables which have column with the same names (ex> id, create_time, update_time) and retrieving the columns, I get unexpected results. The returned rows (maps) have mismatched column names and values!

The reason is that (jasync) `getColumnMetadatas()` [executed in `ColumnMapRowMapper`](https://github.com/spring-projects/spring-framework/blob/19c7527376233c6a5ea42fb6920d004b7be79d66/spring-r2dbc/src/main/java/org/springframework/r2dbc/core/ColumnMapRowMapper.java#L58) in spring-r2dbc returns missing columns.

I debugged some codes, and figured out that this is because of [implementation of `JasyncMetadata`](https://github.com/jasync-sql/jasync-sql/blob/3f5dbaddb6ce8410b198384f10133d0f59cae303/r2dbc-mysql/src/main/java/JasyncMetadata.kt#L40) of like below.

```kotlin
private val metadata: Map<String, ColumnMetadata> = columnNames.map { it to JasyncColumnMetadata(it) }.toMap()
```

Also I checked the implementation of the r2dbc-spi interface [`RowMetadata`](https://github.com/r2dbc/r2dbc-spi/blob/160cd2743f8ce8551618abcd96fa1d7ee37f8f38/r2dbc-spi/src/main/java/io/r2dbc/spi/RowMetadata.java#L54) of the latest version of [r2dbc-mariadb](https://github.com/mariadb-corporation/mariadb-connector-r2dbc) doesn't have this issue. It returned all columns regardless of the same names.

Actually, in jasync-r2dbc-mysql versions compatible with r2dbc-spi versions below 1 (which used `getColumnNames()`), there was no issue like it.

If I'm right, can I try to fix this issue? However, if you think the context related with this issue is too difficult or complex for a new contributor, please tell me.

davin111:
Resolve #390.

Also, I modified the way to build a map of `ColumnMetadata` (using `reversed()`). Because originally [the doc about method implementation (`getColumnMetadata()` and `contains()`)](https://github.com/r2dbc/r2dbc-spi/blob/160cd2743f8ce8551618abcd96fa1d7ee37f8f38/r2dbc-spi/src/main/java/io/r2dbc/spi/RowMetadata.java#L42) of r2dbc interface `RowMetadata` which says `When a get method contains several columns with same name, then the value of the first matching column will be returned` has been potentially violated (**the last** is used, not **the first**). Although this is meaningless because currently the only implemented method of `ColumnMetadata` is `getName()`.

### DIFF
--- a/r2dbc-mysql/src/main/java/JasyncMetadata.kt
+++ b/r2dbc-mysql/src/main/java/JasyncMetadata.kt
@@ -6,6 +6,10 @@ import io.r2dbc.spi.RowMetadata
 import io.r2dbc.spi.Type
 
 class JasyncMetadata(rows: ResultSet) : RowMetadata {
+    private val columnNames = rows.columnNames()
+    private val metadataList = columnNames.map { JasyncColumnMetadata(it) }
+    private val metadataMap = metadataList.reversed().associateBy { it.name }
+
     override fun getColumnMetadata(index: Int): ColumnMetadata {
         if (index >= this.columnNames.size) {
             throw ArrayIndexOutOfBoundsException(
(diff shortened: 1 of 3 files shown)


oshai:
Thanks for the issue and contribution. If possible, consider adding a test similar to: https://github.com/jasync-sql/jasync-sql/blob/master/r2dbc-mysql/src/test/java/com/github/jasync/r2dbc/mysql/integ/JasyncR2dbcIntegTest.kt

Hopefully this will prevent future regression.

davin111:
@oshai Ok, I'll try to add commits for tests soon.

davin111:
@oshai I added a test which failed before this PR and passes now.

oshai:
Thanks for contributing!

oshai:
Releasing it in 2.1.24.

PR #394, opened 2023-06-08
[r2dbc-mysql] Using MySQL default port when not specified

### LINKED ISSUE #393: [r2dbc-mysql] Allowing MySQL port omission
When using r2dbc-mysql, we should specify the db port, like `r2dbc:mysql://10.2.3.4:3306/db`. If we don't specify it,`MysqlConnectionFactoryProvider` insists it doesn't support the url. However, I think allowing port omission is reasonable and convenient. In fact, some other r2dbc drivers allow it and use the default port (`3306`) in that case. Personally, I and some of my colleagues have experienced this minor issue when changing r2dbc driver to jasync-sql.

I'll try to make a PR related this issue, soon.

davin111:
Resolve #393 

Use MySQL default port `3306` when a port to use is not specified.


oshai:
Thanks!

PR #385, opened 2023-03-09
Added missing R2dbc exceptions

pearlgupta2000:
Added handling for R2dbcTimeoutException, R2dbcRollbackException and R2dbcTransientResourceException

Issue #350, opened 2023-01-05
Missing R2dbcExceptions mapping 


pearlgupta2000:
In commit: [https://github.com/jasync-sql/jasync-sql/pull/116/files#diff-48ac27ec5f7ffa2729c94453b580b09eed8a1a59e32d3836d52d83ba0e91d6b5](https://github.com/jasync-sql/jasync-sql/pull/116/files#diff-48ac27ec5f7ffa2729c94453b580b09eed8a1a59e32d3836d52d83ba0e91d6b5), file: r2dbc-mysql/src/main/java/JasyncStatement.kt,

If MySQLException is catched, R2dbcTransientExceptions like R2dbcTimeoutException, R2dbcRollbackException and R2dbcTransientResourceException are not mapped properly because of which each of them falls in JasyncDatabaseException.

Because of which, springframework.r2dbc (org.springframework:spring-r2dbc v5.3.19) always throws them as UncategorizedR2dbcException, whereas in ideal case it should be throwing them as QueryTimeoutException, ConcurrencyFailureException and TransientDataAccessResourceException respectively.

oshai:
Hi, thanks for reporting - would you like to suggest a fix / test for that?

pearlgupta2000:
Yes for fixing this, R2dbcTransientExceptions like like R2dbcTimeoutException, R2dbcRollbackException and R2dbcTransientResourceException should be mapped/handled here : [https://github.com/jasync-sql/jasync-sql/blob/cb42f3f88cadf80e07d6b358c8853e1650482070/r2dbc-mysql/src/main/java/JasyncStatement.kt#L107](https://github.com/jasync-sql/jasync-sql/blob/cb42f3f88cadf80e07d6b358c8853e1650482070/r2dbc-mysql/src/main/java/JasyncStatement.kt#L107).

`errorMessage.errorCode == 3024 -> R2dbcTimeoutException(
                                errorMessage.errorMessage,
                                errorMessage.sqlState,
                                errorMessage.errorCode,
                                throwable
                            )`
Similarly for others like R2dbcRollbackException, R2dbcTransientResourceException, R2dbcNonTransientResourceException and R2dbcDataIntegrityViolationException. 

For errorcodes reference : [https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html](https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html)

[...]

PR #395, opened 2023-06-08
[r2dbc-mysql] Improve MySQL error mapping

davin111:
Related issue: #350 

I referred to other libraries related MySQL and the MySQL official documentation to improve exception mappings. Also, I found #385 PR had been merged but not actually released because the branch was not `master`. This PR covers it, too. Also, I utilized `sqlState` for more complete mapping.


oshai:
Thanks!

PR #418, opened 2023-11-03
Add query to InsufficientParametersException

kennethtiller:
This PR adds query to InsufficientParametersException, to include the SQL in the exception message.

This makes it easier to find what code has failed with insufficient parameters by searching for the query  in client codebase.

It can be hard to find the code because the asynchronous nature of the library does not include stacktrace to call site.


oshai:
Thanks for the PR! can you please add a test for it?

kennethtiller:
> Thanks for the PR! can you please add a test for it?

Thanks for asking! I found it to be a natural fit for the existing tests, have a look and let me know if you'd like it another way

oshai:
lgtm

Issue #289, opened 2022-04-26
mysql IN operator: An operation is not implemented: couldn't find mapping for class java.util.ArrayList


sysmat:
- code
```java
String sql = "SELECT id, name FROM organization WHERE status='ok' AND id IN (?)"
 List<List<Integer>> list = new ArrayList<>();
 list.add(orgsId);
con.sendPreparedStatement(sql, list)
               .get()
               .getRows()
               .forEach(row ->{
                   result.add(new Org(row.getInt(0), row.getString(1)));
                });
```
- error: `An operation is not implemented: couldn't find mapping for class java.util.ArrayList`

oshai:
This is possible in [jdbc](https://stackoverflow.com/questions/17842211/how-to-use-an-arraylist-as-a-prepared-statement-parameter) so we can try to implement it here in the drivers. PR is welcome!

PR #434, opened 2025-06-08
List Support for SQL IN Clause

Sung-Heon:
## Description
This PR adds support for List encoding and resolves mapping issues when using ArrayList with SQL IN clauses.

## Key Changes
Added ListEncoder Support in BinaryRowEncoder - Enables conversion of List types to MySQL length-encoded comma-separated strings
Created ListEncoder Tests - Comprehensive test suite verifying various list encoding scenarios (empty lists, single/multiple items, null filtering, etc.)
Implemented ArrayListInClause Tests - Demonstrates the issue with ArrayList in SQL IN clauses and provides two solutions:
Using correct SQL placeholders

## Related Issue
https://github.com/jasync-sql/jasync-sql/issues/289

### DIFF
--- a/mysql-async/src/main/java/com/github/jasync/sql/db/mysql/binary/BinaryRowEncoder.kt
+++ b/mysql-async/src/main/java/com/github/jasync/sql/db/mysql/binary/BinaryRowEncoder.kt
@@ -13,6 +13,7 @@ import com.github.jasync.sql.db.mysql.binary.encoder.DurationEncoder
 import com.github.jasync.sql.db.mysql.binary.encoder.FloatEncoder
 import com.github.jasync.sql.db.mysql.binary.encoder.IntegerEncoder
 import com.github.jasync.sql.db.mysql.binary.encoder.JavaDateEncoder
+import com.github.jasync.sql.db.mysql.binary.encoder.ListEncoder
 import com.github.jasync.sql.db.mysql.binary.encoder.LocalDateEncoder
 import com.github.jasync.sql.db.mysql.binary.encoder.LocalDateTimeEncoder
 import com.github.jasync.sql.db.mysql.binary.encoder.LocalTimeEncoder
@@ -39,6 +40,7 @@ import java.time.OffsetDateTime
 class BinaryRowEncoder(charset: Charset) {
 
     private val stringEncoder = StringEncoder(charset)
+    private val listEncoder = ListEncoder(charset)
     private val encoders: Map<Class<*>, BinaryEncoder> = mapOf(
         String::class.java to this.stringEncoder,
         BigInteger::class.java to this.stringEncoder,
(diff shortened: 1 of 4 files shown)


Sung-Heon:
@oshai  can you help me with this?
```
Run gradle/wrapper-validation-action@e6e38bacfdf1a337459f332974bb2327a31aaf4b
  with:
    min-wrapper-count: 1
    allow-snapshots: false
  env:
    JAVA_HOME: /opt/hostedtoolcache/Java_Adopt_jdk/8.0.45[2](https://github.com/jasync-sql/jasync-sql/actions/runs/15515837938/job/43682601835#step:4:2)-9/x64
```

What do I have to fix?

oshai:
> @oshai can you help me with this?
> 
> ```
> Run gradle/wrapper-validation-action@e6e38bacfdf1a337459f332974bb2327a31aaf4b
>   with:
>     min-wrapper-count: 1
>     allow-snapshots: false
>   env:
>     JAVA_HOME: /opt/hostedtoolcache/Java_Adopt_jdk/8.0.45[2](https://github.com/jasync-sql/jasync-sql/actions/runs/15515837938/job/43682601835#step:4:2)-9/x64
> ```
> 
> What do I have to fix?

I just rerun it, hth.

Sung-Heon:
@oshai Thx. Please review my code then.😃

oshai:
I guess for the long run it makes more sense to support `in (?)` with list support?

can mysql client/server protocol pass a list object in one placeholder like that?

Sung-Heon:
@oshai  You're right.  In [JDBC](https://docs.oracle.com/javase/6/docs/api/java/sql/PreparedStatement.html#setArray%28int,%20java.sql.Array%29:~:text=ARRAY%20value%20when%20it%20sends%20it%20to%20the%20database.-,Parameters%3A,x%20%2D%20an%20Array%20object%20that%20maps%20an%20SQL%20ARRAY%20value,-Throws%3A)

It supports the SQL ARRAY data type. However, MySQL doesn't have an ARRAY type, unlike PostgreSQL.

So, perhaps I can adapt this code for PostgreSQL?

oshai:
yes, I think postgres will better fit here. for mysql you can see what the jdbc mysql driver does (it still needs to support arrays) and do something similar.

Sung-Heon:
Hi @oshai,

## Proposal: Safe List Expansion for IN Clauses

After careful review, our most robust and safest path forward is to **support list expansion for `IN` clauses exclusively when using named parameters**, across all supported databases. This approach aligns with how [Spring JDBC](https://docs.spring.io/spring-framework/reference/data-access/jdbc/parameter-handling.html#jdbc-in-clause) handles such scenarios.

### Why Automatic Expansion for Positional Placeholders is Unsafe

Attempting to automatically expand positional `?` placeholders is inherently unsafe due to the tight coupling it creates between the Java parameter order and the SQL query's structure.

Consider a query like:

```sql
SET status = ? WHERE id IN (?)
```

If the library were to automatically expand `IN (?)`, it would have to guess the developer's intent based on the parameter order. Should the SQL ever be refactored (e.g., to `WHERE id IN (?) SET status = ?`), a developer would also need to remember to reorder the Java parameters accordingly. This introduces **fragility and makes the code difficult to maintain**.

### The Advantage of Named Parameters

**Named parameters**, such as `:ids`, completely eliminate this risk by removing any dependency on parameter order. This ensures clarity and maintainability.

### Final Implementation Plan

Based on this analysis, our final implementation plan is as follows:

* **For Named Parameter Queries:** We will implement client-side rewriting.
    * For example, `IN (:ids)` will be expanded to `IN (?, ?, ?)` for databases like MySQL.
* **For Positional `?` Queries:** We will **not** implement any automatic expansion. Users will be expected to build the `IN` clause manually. This maintains the standard, predictable, and safe behavior.

This approach allows us to provide a powerful convenience feature—list expansion in `IN` clauses—in a **predictable and safe manner**, without the inherent dangers of adding "magic" to ambiguous placeholders.

---

What are your thoughts on this direction?

oshai:
Currently the driver do not support named params at all. But when using with the r2dbc adapter it add such support iirc.

So I don't know if adding such support is beneficial. It will require us in the client side to parse the statement which we don't do right now (might impct performance, be fragile, etc')

So I would start with just supporting it in postgres by sending array param.

Sung-Heon:
Screenshot 2025-06-19 at 6 25 12 PM

@oshai ok~ I see.

So in postgres does it gonna change like = ANY(array['Ava', 'Dave']) ?

oshai:
Not sure I follow.

